Issue #32, opened 2018-05-15
Add view of the codeflow message inline in the editor


rscrivens:
Add an inline view of the codeflow steps message. This will allow the user to see the message/description right next to the code. 

Open questions:
* do we want to show all of the messages or only the current one selected?
* should we show the message in the line above(via codelens) or to the right of the line(via decoration)?

dgmelski:
Is below the line an option? I think that's what CodeSonar and Clang Static Analyzer do.

Either above or below may sometimes split a multi-line statement. I'm not sure what existing tools do in that case. It might be an argument for "to the right."

Which messages are shown inline could be tied to the "level of detail" option. Arguably should be. (You could have two separate "level of detail" sliders, but that seems a mess.)

In the code view, I'd like the level of detail to control whether messages are displayed or not, but I think involved lines should always be highlighted, regardless of the level of detail.

michaelcfanning:
I believe most viewers puts this information below the line. @rscrivens, VS code lens can provide another data point as far as current viewer behavior.

I tend to agree we could consider two exclusive modes, either the discrete code flow navigator or the inlined view. We definitely should have a single detalis widget that applies to either experience.

We will retain path highlighting for verbose message inlined in code, but the events chrome and other widgets/etc. would disappear. Chrome/etc. for messages hidden due to detail level could still be available on hover.

rscrivens:
After investigating/prototyping, the options for below the region either have issues or aren't supported yet:
- Codelens Below: only seem to display above the region it's associated with, so to get it below I assign the code lens to the line after the end of the region. But this can cause indentation issues if the next line is indented differently then the region the message is for:
![image](https://user-images.githubusercontent.com/34690530/40807716-f1aae212-64d9-11e8-8b13-9c29a37b091d.png)

- View zone widget: VSCode has a widget which looks like it might work well for what we need but I couldn't find anywhere it's currently exposed for extensions to use. There are a few requests for a custom inline widget for extensions to use such as https://github.com/Microsoft/vscode/issues/3220.

For now the best option I can see is using codelens showing the message above region, here's an example of that (includes the functionality to hide/show the message based on verbosity setting):
![inlinecodelens](https://user-images.githubusercontent.com/34690530/40808192-3cb39906-64db-11e8-86bd-997c199de849.gif)

PR #43, opened 2018-06-02
Adds codelens inline messages for the active result's codeflow steps

rscrivens:
Adds codelens inline messages for the active result's codeflow steps, verbosity switch controls which messages show.

Fixes some other minor bugs:
handle undefined step in the message,
handle undefined nestedlevel on call return,
selection highlight was showing on next line if region went to eol
added selection highlight when a result is clicked to highlight the result region


rscrivens:
implements #32

Issue #71, opened 2018-07-24
Rebasing is cumbersome


michaelcfanning:
We lost the file picker, can you please find a way to return it? I was unable to manually update the file path for the SDK SDV tests. The process was cumbersome.

Please follow up with the VS Code team on the failure to provide a mechanism to add a title to the file picker. That is either possible with their API or it is a bug they should fix. It is a significant issue is a file picker comes up with no useful title for users.

rscrivens:
Thanks for the feedback. I'll add this to the backlog to investigate in a future milestone. An initial idea to discuss:

The vscode team has a proposed api that gives an inputbox with more control. It has a way to set buttons, so maybe one option is to keep the current inputbox look and add in a button to pop open an open file picker to it, then once the file is selected we can update the string in the input box with the updated file location.

If we go this route we could also add a button to skip mapping the current file and move to the next one, this would allow a user to selectively decide not to map a file vs. the current behavior where once you hit escape it stops asking the user if it can't map a file(you're still able to manually remap a result via the code action light bulb in the editor).

michaelcfanning:
Makes sense. If VS Code is close to shipping a relevant improvement, let's wait for it.

In the meantime, thanks for your offline comments on the feature. I was confused by the UI because it only verifies the path looks valid. It doesn't actually confirm whether the file exists until you hit enter. We might consider doing the file existence check right away. If it exists on disk, shouldn't we just accept it?

After discovering that the file doesn't exists, I see your guidance about providing a file:// uri. I actually tried to update the file path in the first UI (where this guidance doesn't exist).

michaelcfanning:
Note that uriBaseId use greatly simplifies this problem, you just need to query the user to resolve the set of unique uriBaseIds encountered in the file.

rscrivens:
Regarding the file existence check on initial value, we do set a validator, but vscode handles when to validate and what to do when we return a message for invalid data. When I checked this in it was working correctly where it would show the invalid redbox and message on the initial value. But it looks like vscode made a change that while it still validates the initial value it no longer shows the invalid message and redbox untill a user hit's enter or changes the value: https://github.com/Microsoft/vscode/issues/52908

rscrivens:
[...]

Issue #83, opened 2018-08-28
Source file rebasing textbox doesn't allow normal local file paths


EasyRhinoMSFT:
C:\src\blah.cs should work, but file:/// prefix is required.

rscrivens:
With the latest changes in Dev the user now just needs to enter the "normal" local file path ex. d:\folder\file.ext

PR #85, opened 2018-09-04
Updated user input dialog for remapping

rscrivens:
converted the user input for remapping dialog over to the new more customizable input dialog
•	Fixed the dialog not showing the error box until after you change the value, it now shows it from the start because the file location can’t be found from the start
•	Added two buttons to the new dialog:
o	one lets you open a file picker to select the file and fills the text with the path you selected
o	one lets you skip mapping the file it’s asking for and go to the next file that needs to be remapped
•	Removed the “file:///”  from being displayed in the input dialog and from being required. Now the user only needs to input a “normal” path like d:\folder\file.ext


rscrivens:
Addresses/Fixes #71 & #83

Issue #81, opened 2018-08-28
Code flow items get ugly with a long file path


EasyRhinoMSFT:
Causes the location text to wrap. Suggest ellipsis truncation if possible (but keep the line # visible)

rscrivens:
good point, I'll look into adding a truncation for the file name to keep it cleaner.

rscrivens:
Added an ellipse, and tweaked the layout to better show the code flows with larger file names and when the window width is reduced

PR #87, opened 2018-09-05
Tweaks to the file name and line number for better viewing when the s…

rscrivens:
Changes the file name to show as ellispes if the window is shrunken and the file name is too long. keeps the line number. Also changed the word break to break all to help readability.


rscrivens:
fixes for #81

PR #94, opened 2018-09-15
A couple of fixes, when the file location, rule id, or rule name are not defined

rscrivens:
Adds default values of "No Rule Info" and "No Location" in cases where the values aren't defined. Also if no rule id is defined we don't add that info to the message that shows up in the problems panel.

### DIFF
--- a/src/explorer/webview.ts
+++ b/src/explorer/webview.ts
@@ -211,14 +211,21 @@ class ExplorerWebview {
     private createExplorerHeaderContent(resultInfo: ResultInfo): HTMLDivElement {
         const header = this.createElement("div", { id: "title" }) as HTMLDivElement;
 
-        header.appendChild(this.createElement("label", { id: "titleruleid", text: resultInfo.ruleId }));
-        header.appendChild(this.createElement("label", { id: "titlerulename", text: resultInfo.ruleName }));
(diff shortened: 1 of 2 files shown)


src/explorer/webview.ts
```diff
@@ -211,15 +211,22 @@ class ExplorerWebview {
...
+            header.appendChild(this.createElement("label", { id: "titlerulename", text: resultInfo.ruleName }));
+        } else {
+            header.appendChild(this.createElement("label", { id: "titlerulename", text: "No Rule Info" }));
+        }
         header.appendChild(this.createElement("label", { text: " | " }));
-        if (resultInfo.locations[0] !== undefined) {
-            const filenameandline = resultInfo.locations[0].fileName + " (" +
+
+        let filenameandline = "No Location";
+        if (resultInfo.locations[0] !== null && resultInfo.locations[0] !== undefined) {
```
michaelcfanning: >resultInfo.locations [](start = 12, length = 20)

can resultInfo.Locations itself ever be null?

rscrivens: good question, no the location will always be defined as at least an empty array

---
In reply to: [217875015](https://github.com/Microsoft/sarif-vscode-extension/pull/94#discussion_r217875015) [](ancestors = 217875015)

Issue #129, opened 2019-02-07
Update Sarif version to 2.0.0-csd.2.beta.2019-01-09


rscrivens:
Need to update to the latest sarif version 2.0.0-csd.2.beta.2019-01-09. areas that need to be updated:
* Types - will need to update in definitelytyped then pull down the latest npm
* MultiTool

PR #131, opened 2019-02-12
Update sarif version to 1-09-19

rscrivens:
updated to the 1-09-19 npm type definitions
updated the multitool to the latest
fixed a bug in the generate temp directory


rscrivens:
implements #129

Issue #14, opened 2018-04-24
Add a link in the Result Info tab to the Sarif file


rscrivens:
Add a link in the Result Info that maps back to the location in the Sarif file. This will allow the user to look at the raw sarif results.

michaelcfanning:
Useful debugging feature. Rough equivalent exists in VS add-in. Add at your discretion @rscrivens

PR #140, opened 2019-02-21
Adds link to the results location in the Sarif file to the results info panel in the sarif explorer.

rscrivens:
Also Fixed a bug when trying to delete a file if the sarif explorer opened on top of it, which caused it to be undefined and triggered an error when we try to close undefined.


rscrivens:
Implements #14

PR #196, opened 2019-08-20
Fix for exception when result contains embedded file contents with lo…

cfaucon:
Fix for exception when result contains embedded file contents with location uri that contains illegal path characters (?:, etc).

Example: result location begins with https:// and the SARIF contains embedded file contents for this result artifact. Temporarily file to be generated from the embedded contents attempts to use partial of the location uri which contains illegal path and file name characters.  Adding sanitize method to remove such illegal characters when forming the temporary file path.

### DIFF
--- a/src/Utilities.ts
+++ b/src/Utilities.ts
@@ -206,12 +206,21 @@ export class Utilities {
     public static generateTempPath(filePath: string, hashValue?: string): string {
         const pathObj = Utilities.Path.parse(filePath);
         let tempPath: string = Utilities.Path.join(Utilities.SarifViewerTempDir, hashValue || "",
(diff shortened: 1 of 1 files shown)


src/Utilities.ts
```diff
@@ -206,14 +206,23 @@ export class Utilities {
     public static generateTempPath(filePath: string, hashValue?: string): string {
```
michaelcfanning: >generateTempPath [](start = 18, length = 16)

Will this code handle illegal chars in the directory name? Such as escaped spaces? 

[https://element/%20%subelementhello%20%/blah.cpp](https://element/%20%subelementhello%20%/blah.cpp)

PR #201, opened 2019-09-17
Fix #200: File open dialog for conversion restricts filename extension

### LINKED ISSUE #200: File open dialog for conversion restricts filename extension
1. Press F1. The Command Palette opens.

2. Type `sarif:Convert and open a non-sarif file`. (It will auto-complete after a few characters). A list of supported tools appears.

3. Select `ClangAnalyzer`. The Open File dialog opens.

image

Note that the filter dropdown only offers one option: `ClangAnalyzer log files (*.xml)`. It should also have an option `All files (*.*)` in case the file has a different extension, like `.plist`.

You _can_ open files with other extensions, but you have to type the filename by hand because it won't appear in the files list.

I saw this on Windows. I don't know if it happens on Linux.

I haven't checked to see if the Open File dialogs for the other tools have the same problem, but we should make sure they all offer `*.*`.

@LambdaScientist-1stEdge @michaelcfanning FYI

lgolding:
This is the same issue as microsoft/sarif-visualstudio-extension#104 that I fixed yesterday. The customer @LambdaScientist-1stEdge actually reported it here, in the VS Code extension.


lgolding:
@michaelcfanning is added to the review. #Closed

lgolding:
@rtaket is added to the review. #Closed

lgolding:
@cfaucon is added to the review. #Closed

lgolding:
@harleenkohli is added to the review. #Closed

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Change Log
 
+## 2.16.0
```
michaelcfanning: This is the first time I've noticed the the VS code extension lists most recent work at the top rather than the bottom of the file. :) We should consider this for other repos.

src/FileConverter.ts
```diff
@@ -31,6 +31,7 @@ export class FileConverter {
             const dialogOptions = { canSelectFiles: true, canSelectMany: false, filters: {} } as OpenDialogOptions;
             const toolExt = FileConverter.Tools.get(tool);
             dialogOptions.filters[`${tool} log files`] = toolExt;
+            dialogOptions.filters['All files'] = [ '*' ]
```
michaelcfanning: >*' [](start = 52, length = 2)

Why is this a good idea? Can't believe I didn't mention this earlier. Tools don't produce output with arbitrary extensions. If you provide a filter for all files, there's no reason to show files of any particular extension (as they will already show). This eliminates that value of, say, reducing a directory with hundreds of files to the subset that are actually relevant to, e.g, a microfocus conversion. What I would have expected is that we added the relevant additional extension for Pylint over showing all files. How does showing a docx file assist a user for any conversion scenario?

lgolding: >  If you provide a filter for all files, there's no reason to show files of any particular extension (as they will already show).

That's not true. If shows the files for the first filter in the list. It only shows "All files" if you select it from the filter drop-down. Do you still object to the "*"?

---
In reply to: [325719497](https://github.com/microsoft/sarif-vscode-extension/pull/201#discussion_r325719497) [](ancestors = 325719497)

CHANGELOG.md
lgolding: Yes! I'm going to raise PRs on other repos to fix this. I'll separate them from all other PRs.

---
In reply to: [325717773](https://github.com/microsoft/sarif-vscode-extension/pull/201#discussion_r325717773) [](ancestors = 325717773)

src/FileConverter.ts
lgolding: In addition to the "*", I can add ".plist" to the Fortify extension list, and (for that matter ".sarif.json" to the SARIF extension list. That addresses the customer's spot problem but I still think "*" is good to have for customers who for whatever reason have files with different extensions.

---
In reply to: [325887958](https://github.com/microsoft/sarif-vscode-extension/pull/201#discussion_r325887958) [](ancestors = 325887958,325719497)

michaelcfanning: Yes, let's do what you describe and close this PR. :)

---
In reply to: [325893269](https://github.com/microsoft/sarif-vscode-extension/pull/201#discussion_r325893269) [](ancestors = 325893269,325887958,325719497)

jeffersonking: @lgolding: You want `sarif.json` over plain `json`?

lgolding: That's a good point. Let's make it plain `.json`. That will pick up any `.sarif.json` files, and I don't see value in excluding `.json` files that _aren't_ `.sarif.json`.

---
In reply to: [406450470](https://github.com/microsoft/sarif-vscode-extension/pull/201#discussion_r406450470) [](ancestors = 406450470)

jeffersonking:
Incorporated Larry's suggestion of "*.plist" for Fortify files also as part of the "merge".

src/FileConverter.ts
jeffersonking: Discussed with Larry offline that `*.json` actually does not apply in the scenario. Thus omitting this aspect of the suggestion.

PR #208, opened 2020-03-06
Use import for 'fs' and 'path' rather than require, and fix casing of paths from SARIF results.

GabeDeBacker:
This PR changes from the "node" technique of "Require" to the typescript "import" for path and fs, which makes VSCode's IntelliSense function during development. The modules are already loaded by VSCode way before the SARIF extension even loads.

Next up is fixing the path casing from the SARIF result so VSCode doesn't open multiple editors for the same files as VSCode treats URIs as case sensitive. The technique involves enumerating the directory entries path fragment by path fragment to retrieve the proper casing from the file system. While this sounds expensive, on Windows I can confirm that this operation takes mere milliseconds. The ultimate goal would be to have compilers, analyzers, etc. to not mess around with the casing of file names and paths, but that might likely be an impossible goal :(

The last minor change was to have the "build" task for the VSCode extension actually execute the compilation of the explorer web-view. It took me a while to figure out why everything in the extension was working BUT the web-view :) So I addressed that so others would not be confused like I was. :)


src/Utilities.ts
```diff
@@ -327,32 +308,66 @@ export class Utilities {
...
+        const tempPath = path.join(Utilities.Os.tmpdir(), Utilities.SarifViewerTempDir);
+        Utilities.removeDirectoryContents(tempPath);
+        fs.rmdirSync(tempPath);
+    }
+
+    /**
+     * Fixes up file schemed based paths to contain the proper casing for VSCode's URIs which are case sensitive.
+     * @param uri The URI for which to fix the casing.
+     */
+    public static fixUriCasing(uri: Uri): Uri {
```
michaelcfanning: >fixUriCasing [](start = 18, length = 12)

Perfect method for some sort of unit-testing. Does this code base even establish a pattern for that? :P Jeff/Larry, do you know?

src/Utilities.ts
```diff
@@ -327,32 +308,66 @@ export class Utilities {
...
+        const pathParts: string[] = uri.fsPath.split(path.sep);
+        if (pathParts.length < 1) {
+            return uri;
+        }
+
+        // We assume that the "root" drive is lower-cased which is "usually" the case.
+        const fixedPathParts: string[] = [ pathParts[0].toLowerCase() + path.sep];
+
+        while (pathPartIndex + 1 < pathParts.length) {
+            const directoryToRead: string = path.join(...fixedPathParts);
```
michaelcfanning: >path.join [](start = 44, length = 9)

You could replace all joins by building up a 'fixedPath' variable by concat'ing new parts as you encounter them (I assume each iteration of this single append would be more efficient than the join). I don't think this would compromise readability too much. I doubt the micro-optimization has measurable impact but if you agree it doesn't compromise readability, something to consider. Our pathological case here on windows is roughly MAX_PATH/2 (c:\a\b\c\etc.), so somewhere around ~130 parts. in C#, we could use the directory info object to avoid building out all the intermediate paths, not sufficiently knowledgeable here to know whether you have a TS/JS alternate...

src/Utilities.ts
```diff
@@ -327,32 +308,66 @@ export class Utilities {
...
+        fs.rmdirSync(tempPath);
+    }
+
+    /**
+     * Fixes up file schemed based paths to contain the proper casing for VSCode's URIs which are case sensitive.
+     * @param uri The URI for which to fix the casing.
+     */
+    public static fixUriCasing(uri: Uri): Uri {
+        // We can only support file-system scheme files.
+        if (uri.scheme.toUpperCase() !== "FILE") {
```
lgolding: >toUpperCase [](start = 23, length = 11)

Is there a `toUpperInvariant`? This upper-casing of the `file` scheme was the source of the notorious "Turkish i" bug.

src/Utilities.ts
```diff
@@ -327,32 +308,66 @@ export class Utilities {
...
+            return uri;
+        }
+
+        let pathPartIndex: number = 0;
+        const pathParts: string[] = uri.fsPath.split(path.sep);
+        if (pathParts.length < 1) {
+            return uri;
+        }
+
+        // We assume that the "root" drive is lower-cased which is "usually" the case.
```
lgolding: >. [](start = 85, length = 1)

I understand that by lower-casing the drive, you avoid duplicates ("[file:///C:/file.c](file:///C:/file.c)" vs. "[file:///c:/file.c](file:///c:/file.c)"). So I agree with the line of code. But is the comment right? In what was does forcing the drive letter to lower case embody an assumption?

src/Utilities.ts
michaelcfanning: Couldn't hurt, though I'm not sure about adverse impact of a bad comparison. If you had files that collided in this way, we might open the wrong thing. This is no different than what would happen with any file duplication where elements differed by casing only. So, we have the same problem already (with no path to a perfect solution).

---
In reply to: [389079043](https://github.com/microsoft/sarif-vscode-extension/pull/208#discussion_r389079043) [](ancestors = 389079043)

PR #210, opened 2020-03-15
Fix #204

### LINKED ISSUE #204: Unable to scroll inside the SARIF Explorer tab (Results list is too big)
Hi, 

I have over 2900 results in the Results List in the SARIF Explorer tab, so the list overflows as expected with so many results. But there is no scrollbar in the SARIF Explorer view, using the mouse scroller does not have any effect eider and keyboard navigation doesn't seem to be supported eider. 

I don't see any exception messages showing up anywhere during the parsing of the sarif file, so I assume that the plugin is working correctly for most of it.

Is there some way to be enable scrolling in the SARIF Explorer view? If this is a bug how can I help out to debug this? (As I don't see any exceptions showing up anywhere)

jeffersonking:
Overflow hidden on #resultslistcontainer, causes overflow scroll on #resultslisttablecontainer to take effect.

### DIFF
--- a/resources/explorer/resultsList.css
+++ b/resources/explorer/resultsList.css
@@ -14,6 +14,7 @@
     flex: 1 1 50vh;
     flex-flow: column nowrap;
     padding: 5px;
+    overflow: hidden;
 }
 
 #resultslisttablecontainer {


jeffersonking:
Finishing up the work Wilson Wong @wwong started here: https://github.com/microsoft/sarif-vscode-extension/pull/209/commits

PR #211, opened 2020-03-15
Turn on lint rules and up type-script compiler warnings

GabeDeBacker:
In order to ensure that during extension development we do not make simple coding mistakes (such as not waiting for promises, not handing undefined\null, etc.) I have turned on lint rules and typescript compiler options to help catch these issues.

While doing so, I obviously had to touch a lot (a LOT) of code to fix all the warnings. I also had to correct code that simply was not handling undefined\null properly.

I need everyone's help verifying the extension.


michaelcfanning:
Gabe, this change appears to break the viewer for files that don't have file locations (such as our PE checker that only flags problems at the binary level).

image

jeffersonking:
I'm following your new code comments. Very useful as I'm ramping up.

src/ExplorerController.ts
```diff
@@ -66,87 +119,115 @@ export class ExplorerController {
...
                     uri: Uri.parse(locData.file),
-                } as Location;
-                CodeFlowDecorations.updateSelectionHighlight(location, undefined);
+                    toJSON: Utilities.LocationToJson
+                };
+                await CodeFlowDecorations.updateSelectionHighlight(this,  location, undefined);
                 break;
-            case MessageType.VerbosityChanged:
+
+                case MessageType.VerbosityChanged:
```
jeffersonking: Indentation.

src/ExplorerController.ts
```diff
@@ -156,13 +237,12 @@ export class ExplorerController {
...
-    public setActiveDiagnostic(diag: SarifViewerDiagnostic, mappingUpdate?: boolean) {
-        if (this.activeSVDiagnostic === undefined || this.activeSVDiagnostic !== diag || mappingUpdate) {
-            this.activeSVDiagnostic = diag;
+    public setActiveDiagnostic(diag: SarifViewerVsCodeDiagnostic, mappingUpdate?: boolean): void {
+        if (!this.activeDiagnostic || this.activeDiagnostic !== diag || mappingUpdate) {
+            this.activeDiagnostic = diag;
             if (!mappingUpdate) {
                 this.activeTab = undefined;
-                this.selectedRow = undefined;
-                this.selectedVerbosity = undefined;
```
jeffersonking: Verbosity no longer resets on mapping update?

src/FileConverter.ts
```diff
@@ -93,111 +125,135 @@ export class FileConverter {
      * @param sarifVersion version of the sarif log
      * @param sarifSchema version of the sarif logs schema
      */
-    public static async upgradeSarif(doc: TextDocument, sarifVersion?: SarifVersion, sarifSchema?: SarifVersion) {
-        const saveTemp = "Yes (Save Temp)";
-        const saveAs = "Yes (Save As)";
-        const msgOptions = { modal: false } as MessageOptions;
+    public static async upgradeSarif(doc: vscode.TextDocument, sarifVersion?: SarifVersion, sarifSchema?: SarifVersion): Promise<boolean> {
+        interface UpgradeChoiceMessageItem extends vscode.MessageItem {
```
jeffersonking: No longer required.

src/extension.ts
```diff
@@ -1,44 +1,37 @@
...
+    Utilities.initialize(context);
+    FileConverter.initialize(context);
 
-    context.subscriptions.push(
-        commands.registerCommand(ExplorerController.SendCFSelectionToExplorerCommand, (id: string) => {
-            CodeFlowDecorations.updateCodeFlowSelection(id);
-            ExplorerController.Instance.setSelectedCodeFlow(id);
-        }));
+    const explorerController: ExplorerController = new ExplorerController(context);
+    context.subscriptions.push(explorerController);
```
jeffersonking: We could do ```
context.subscriptions.push(
    explorerController,
    codeActionProvider,
    ....and so on...
)
```

File: src/extension.ts [hunk 5]
```diff
@@ -47,26 +40,19 @@ export function activate(context: ExtensionContext) {
...
-    context.subscriptions.push(CodeFlowCodeLensProvider.Instance);
-    context.subscriptions.push(FileMapper.Instance);
-
     // Read the initial set of open SARIF files
-    const reader = LogReader.Instance;
+    const reader: LogReader = (new LogReader(explorerController));
     context.subscriptions.push(reader);
+
+    // We do not need to block extension startup for reading any open documents.
+    // tslint:disable-next-line: no-floating-promises
```
jeffersonking: Common to use `void reader.readAll()` to express such an intent.

src/explorer/webview.ts
```diff
@@ -137,59 +147,46 @@ class ExplorerWebview {
...
      * Cleans up the result details section, removing event handlers and elements
      */
-    private cleanUpResultDetails() {
+    private cleanUpResultDetails(): void {
         // Remove event handlers
-        const tabContainer = document.getElementById("tabcontainer");
-        if (tabContainer !== null) {
-            for (let i = 0; i < tabContainer.children.length; i++) {
-                tabContainer.children.item(i).removeEventListener("click", this.onTabClickedBind);
+        const tabContainer: HTMLDivElement | undefined = getOptionalDocumentElementById(document, "tabcontainer", HTMLDivElement);
```
jeffersonking: What is the intent of down casting to `HTMLDivElement` here?

src/explorer/webview.ts
```diff
@@ -137,59 +147,46 @@ class ExplorerWebview {
...
-        while (header.children.length > 0) {
-            header.removeChild(header.children.item(0));
-        }
-
-        const container = document.getElementById("resultdetailscontainer");
-        while (container.children.length > 0) {
-            container.removeChild(container.children.item(0));
-        }
+        removeElementChildren(document.getElementById("resultdetailsheader"));
+        removeElementChildren(document.getElementById("resultdetailscontainer"));
```
jeffersonking: I'm unclear on the purpose of `removeElementChildren()` over the simpler `document.getElementById("resultdetailscontainer").innerHTML = ""`?

src/explorer/webview.ts
```diff
@@ -729,104 +738,105 @@ class ExplorerWebview {
...
 
-                    const fLine = fLocation.range[0].line.toString();
-                    const fParameters = frame.parameters.toString();
-                    let fMsg = "";
-                    if (frame.message !== undefined && frame.message.text !== undefined) {
-                        fMsg = frame.message.text;
-                    }
+                const fRow: HTMLTableRowElement = this.createElement("tr", {
+                    attributes: {
+                        "data-eCol": fLocation.range !== undefined ? fLocation.range.end.character.toString() : 0,
```
jeffersonking: Another opportunity for reduction:`fLocation.range?.end.character.toString() ?? 0`.

src/explorer/webview.ts
```diff
@@ -1058,29 +1093,29 @@ class ExplorerWebview {
...
     }
 
     /**
      * Callback when user clicks on the Fix tree
      * @param event event fired when user clicked the fix tree
      */
-    private onFixClicked(event: MouseEvent) {
-        let ele = event.srcElement as HTMLElement;
+    private onFixClicked(event: MouseEvent): void {
+        let ele: HTMLElement | null = <HTMLElement>event.srcElement;
```
jeffersonking: Reason for the switch away from `as` to `<>` in the several of the lines here and below?

src/explorer/webview.ts
```diff
@@ -1,66 +1,75 @@
-// /********************************************************
-// *                                                       *
-// *   Copyright (C) Microsoft. All rights reserved.       *
-// *                                                       *
-// ********************************************************/
+/*!
```
jeffersonking: This legacy DOM construction/manipulation strategy is at its scaling limit and is not time/cost-effective. Recommend avoiding further investment (such as refactoring). This includes time spent reviewing. If the intent is to add major features here, then we should discuss strategy.

src/ExplorerController.ts
GabeDeBacker: This is just a functional bug. The level of verbosity is not intended to change as you move between individual diagnostics.

---
In reply to: [397533596](https://github.com/microsoft/sarif-vscode-extension/pull/211#discussion_r397533596) [](ancestors = 397533596)

src/extension.ts
GabeDeBacker: We discussed why this isn't a great idea over the Teams\Skype\Zoom

---
In reply to: [397540093](https://github.com/microsoft/sarif-vscode-extension/pull/211#discussion_r397540093) [](ancestors = 397540093)

src/explorer/webview.ts
GabeDeBacker: I couldn't agree more. Not super worth it in this PR.

---
In reply to: [397721726](https://github.com/microsoft/sarif-vscode-extension/pull/211#discussion_r397721726) [](ancestors = 397721726)

src/explorer/webview.ts
GabeDeBacker: Actually, after ensuring that a range is always created (even if it is the first line), the undefined check is no longer needed.

---
In reply to: [397704938](https://github.com/microsoft/sarif-vscode-extension/pull/211#discussion_r397704938) [](ancestors = 397704938)

src/explorer/webview.ts
GabeDeBacker: I really want all the "Casts" to go away. The idea (which I have implemented) is a "type safe way" to make sure the cast is of the correct type.

---
In reply to: [397709904](https://github.com/microsoft/sarif-vscode-extension/pull/211#discussion_r397709904) [](ancestors = 397709904)

src/explorer/webview.ts
GabeDeBacker: With struct null checks turned on, 
document.getElementById("resultdetailscontainer").innerHTML = "".

is not valid as getElementById can return null.

And optional property access
document.getElementById("resultdetailscontainer")?.innerHTML = ""

isn't allowed.

So... the best you can do is have removeElementChildren helper do what you suggest and have it take Element | null.



---
In reply to: [397701937](https://github.com/microsoft/sarif-vscode-extension/pull/211#discussion_r397701937) [](ancestors = 397701937)

src/explorer/webview.ts
GabeDeBacker: The idea is to have type-script code actually ensure that element type you are working on is really what you think it is. But yes, in the same regards, this code should be using the least contract it needs to do it's work.

Changing to HTMLElement.

---
In reply to: [397699182](https://github.com/microsoft/sarif-vscode-extension/pull/211#discussion_r397699182) [](ancestors = 397699182)

src/CodeFlowDecorations.ts
```diff
@@ -188,60 +233,90 @@ export class CodeFlowDecorations {
      * @param idText text version of the id of the Code Flow, set to undefined if using id
      * @param idCFStep Id object of the Code Flow, set to undefined if using idText
      */
-    public static updateCodeFlowSelection(idText?: string, idCFStep?: CodeFlowStepId) {
-        let id: CodeFlowStepId;
-        if (idText !== undefined) {
+    public static async updateCodeFlowSelection(explorerController: ExplorerController, idText?: string, idCFStep?: CodeFlowStepId): Promise<void> {
```
GabeDeBacker: > public static async updateCodeFlowSelection( [](start = 3, length = 45)

These static methods do not need to exist (here and below).

Make the contract between the "Explorer controller" and everything else more clear.

src/CodeFlows.ts
```diff
@@ -75,18 +72,20 @@ export class CodeFlows {
...
-                const thread = codeFlow.threads[tFKey];
+                const thread: ThreadFlow = codeFlow.threads[tFKey];
                 for (const stepKey of thread.steps.keys()) {
-                    const step = thread.steps[stepKey];
-                    if (step.location !== undefined && step.location.mapped !== true) {
-                        const sarifLoc = sarifCodeFlows[cFKey].threadFlows[tFKey].locations[stepKey].location;
-                        await LocationFactory.create(sarifLoc, runId).then((location: Location) => {
-                            codeFlows[cFKey].threads[tFKey].steps[stepKey].location = location;
-                        });
+                    const step: CodeFlowStep = thread.steps[stepKey];
```
GabeDeBacker: Don't iterate the keys and then just index into the object.

GabeDeBacker:
>     /**

This file is really a namespace, not a class...

---
Refers to: src/CodeFlows.ts:16 in a7cd129. [](commit_id = a7cd129044e9533ce40ccfc3ac929200791603c9, deletion_comment = False)

src/FileConverter.ts
```diff
@@ -1,43 +1,67 @@
...
+            canSelectFiles: true,
+            canSelectMany: false,
+            filters
         });
+
+        if (!openUris) {
+            return;
+        }
+        if (openUris.length > 0) {
+            FileConverter.convert(openUris[0], tool.label);
```
GabeDeBacker: >label [](start = 52, length = 5)

Comment the fact that the tool name is the same as the tool-label and that it is passed straight through as an argument to the SDK converter.

src/FileConverter.ts
```diff
@@ -1,43 +1,67 @@
...
+        const openUris: vscode.Uri[] | undefined = await vscode.window.showOpenDialog({
+            canSelectFiles: true,
+            canSelectMany: false,
+            filters
         });
+
+        if (!openUris) {
+            return;
+        }
+        if (openUris.length > 0) {
```
GabeDeBacker: Combine this check

GabeDeBacker:
>     }

This does not need to be a map, or created dynamically.

---
Refers to: src/FileConverter.ts:280 in a7cd129. [](commit_id = a7cd129044e9533ce40ccfc3ac929200791603c9, deletion_comment = False)

src/LocationFactory.ts
```diff
@@ -1,67 +1,79 @@
...
+import { Location, Message, JsonMapping, JsonPointer } from "./common/Interfaces";
 import { LogReader } from "./LogReader";
 import { Utilities } from "./Utilities";
+import { ExplorerController } from "./ExplorerController";
 
 /**
- * Class that processes and creates a location object from a results physicallocation
+ * Class that processes and creates a location object from a results physical location
  */
 export class LocationFactory {
```
GabeDeBacker: >class [](start = 7, length = 5)

Also a namespace and not a class

src/ResultInfoFactory.ts
```diff
@@ -1,195 +1,203 @@
...
 import { Utilities } from "./Utilities";
+import { ExplorerController } from "./ExplorerController";
 
 /**
- * Class that holds the result information processed from the Sarif result
+ * Class that holds the result information processed from the Sarif result.
+ * This code transofrms the SARIF JSON into a "flatter" data model that is used
+ * by the rest of the viewer code.
  */
 export class ResultInfoFactory {
```
GabeDeBacker: >class [](start = 7, length = 5)

also a namespace

GabeDeBacker:
> export class RunInfoFactory {

also a namespace


---
Refers to: src/RunInfoFactory.ts:13 in a7cd129. [](commit_id = a7cd129044e9533ce40ccfc3ac929200791603c9, deletion_comment = False)

PR #214, opened 2020-03-31
Decouple file-mapper from explorer controller and diagnostic collection

GabeDeBacker:
This change performs more decoupling of the UI from the object model by decoupling the file-mapper from being owned by the diagnostic collection (which is then owned by the explorer controller).

Verified by unit-tests and manual testing.

### DIFF
--- a/src/CodeFlowDecorations.ts
+++ b/src/CodeFlowDecorations.ts
@@ -9,7 +9,7 @@ import {
     commands, DecorationInstanceRenderOptions, DecorationOptions, DecorationRangeBehavior, DiagnosticSeverity, OverviewRulerLane,
     Position, Range, TextEditor, TextEditorDecorationType, TextEditorRevealType, Uri, ViewColumn, window, workspace, TextDocument, Disposable,
 } from "vscode";
-import { CodeFlowStep, CodeFlowStepId, Location, CodeFlow, SarifViewerDiagnostic, WebviewMessage, LocationData } from "./common/Interfaces";
+import { CodeFlowStep, CodeFlowStepId, Location, CodeFlow, WebviewMessage, LocationData } from "./common/Interfaces";
 import { ExplorerController } from "./ExplorerController";
 import { Utilities } from "./Utilities";
 import { SarifViewerVsCodeDiagnostic } from "./SarifViewerDiagnostic";
@@ -163,7 +163,7 @@ export class CodeFlowDecorations implements Disposable {
      * Selects the next CodeFlow step
      */
     private async selectNextCFStep(): Promise<void>  {
-        const diagnostic: SarifViewerDiagnostic | undefined = this.explorerController.activeDiagnostic;
+        const diagnostic: SarifViewerVsCodeDiagnostic | undefined = this.explorerController.activeDiagnostic;
         if (!diagnostic) {
             return;
         }
@@ -199,7 +199,7 @@ export class CodeFlowDecorations implements Disposable {
      * Selects the previous CodeFlow step
      */
     private async selectPrevCFStep(): Promise<void> {
-        const diagnostic: SarifViewerDiagnostic | undefined = this.explorerController.activeDiagnostic;
+        const diagnostic: SarifViewerVsCodeDiagnostic | undefined = this.explorerController.activeDiagnostic;
         if (!diagnostic) {
             return;
         }
@@ -249,7 +249,7 @@ export class CodeFlowDecorations implements Disposable {
 
         this.lastCodeFlowSelected = id;
 
-        const diagnostic: SarifViewerDiagnostic | undefined = this.explorerController.activeDiagnostic;
+        const diagnostic: SarifViewerVsCodeDiagnostic | undefined = this.explorerController.activeDiagnostic;
         if (!diagnostic) {
             return;
         }
@@ -286,7 +286,7 @@ export class CodeFlowDecorations implements Disposable {
      * @param sarifLocation raw sarif location used if location isn't mapped to get the user to try to map
      */
     private async updateSelectionHighlight(location: Location, sarifLocation?: sarif.Location): Promise<void> {
-        const diagnostic: SarifViewerDiagnostic | undefined = this.explorerController.activeDiagnostic;
+        const diagnostic: SarifViewerVsCodeDiagnostic | undefined = this.explorerController.activeDiagnostic;
 
         if (!diagnostic) {
             return;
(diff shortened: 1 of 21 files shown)


src/test/index.ts
```diff
@@ -13,7 +13,7 @@ export function run(): Promise<void> {
         fullStackTrace: true,
 
         // Do not allow uncaught exceptions.
-        allowUncaught: false
+        allowUncaught: false,
```
GabeDeBacker: >         [](start = 29, length = 8)

Undo this. Not needed

src/factories/CodeFlowFactory.ts
```diff
@@ -77,14 +77,14 @@ export namespace CodeFlowFactory {
      * @param sarifCodeFlows Used if a codeflow needs to be remapped
      * @param runId used for mapping uribaseids
      */
-    export async function tryRemapCodeFlows(explorerController: ExplorerController, codeFlows: CodeFlow[], sarifCodeFlows: sarif.CodeFlow[], runId: number): Promise<void> {
+    export async function tryRemapCodeFlows(fileMapper: FileMapper, runInfo: RunInfo, codeFlows: CodeFlow[], sarifCodeFlows: sarif.CodeFlow[], runId: number): Promise<void> {
         for (const [cFKey, codeFlow] of codeFlows.entries()) {
             for (const [tFKey, threadFlow] of codeFlow.threads.entries()) {
                 for (const [stepKey, step] of threadFlow.steps.entries()) {
-                    if (step.location && step.location.mapped) {
+                    if (step.location && !step.location.mapped) {
```
GabeDeBacker: >!step.location.mapped [](start = 41, length = 21)

Yes, this was a regression I caught and verified manually as well as the original code (which I kept a copy of)

src/CodeFlowDecorations.ts
```diff
@@ -480,7 +479,7 @@ export class CodeFlowDecorations implements Disposable {
                     uri: Uri.parse(locData.file),
                     toJSON: Utilities.LocationToJson
                 };
-                await this.updateSelectionHighlight(location, undefined);
+                await this.updateSelectionHighlight(this.activeDiagnostic.runInfo, location, undefined);
```
GabeDeBacker: >this.activeDiagnostic.runInfo,  [](start = 52, length = 31)

this function has access to the active diagnostic, so why am I passing it?

src/FileMapper.ts
```diff
@@ -245,7 +245,7 @@ export class FileMapper implements Disposable {
      * @param artifacts array of sarif.Artifact that needs to be mapped
      * @param runId id of the run these files are from
      */
```
GabeDeBacker: I need to make a pass through all of the "public" methods and fix the documentation. Especially one's i've changed.

src/SVDiagnosticCollection.ts
```diff
@@ -182,51 +175,26 @@ export class SVDiagnosticCollection implements Disposable {
...
-                    if (remapped) {
-                        this.add(diag);
-                        this.diagnosticCollectionChangedEventEmitter.fire({
-                            diagnostics: [diag],
-                            type: 'Add'
-                        });
-                        if (explorerDiag !== undefined && explorerDiag.resultInfo.runId === diag.resultInfo.runId &&
-                            explorerDiag.resultInfo.id === diag.resultInfo.id) {
-                            this.explorerController.setActiveDiagnostic(diag, true);
-                        }
```
GabeDeBacker: Couple of comments for this change. First, the "add" function already fires the event, so no need to fire it twice. Second, setting the active diagnostic in the explorer view in a loop is actually not good as if the remapping changes multiple diagnostics, the loop will continue to reset it.

We will need to think about whether a successful mapping should change the active diagnostic in the web-view in the first place.

src/SVDiagnosticCollection.ts
```diff
@@ -182,51 +175,26 @@ export class SVDiagnosticCollection implements Disposable {
...
+                    this.add(unmappedIssue);
+                } else {
+                    remainingUnmappedIssues.push(unmappedIssue);
+                }
             }
 
             if (remainingUnmappedIssues.length === 0) {
                 this.unmappedIssuesCollection.delete(key);
-            } else if (remainingUnmappedIssues.length !== issues.length) {
+            } else if (remainingUnmappedIssues.length !== unmappedIssues.length) {
```
GabeDeBacker: >(remainingUnmappedIssues.length !== unmappedIssues.length)  [](start = 22, length = 59)

This check isn't actually necessary since setting the value is harmless if nothing has changed, and it is actually safer in case for some reason the logic above changes and the items change but the length stays the same.

GabeDeBacker:
Yes. Indeed. But the coupling is decreased by several levels as it isn't buried inside the diagnostic collection which was then buried in the explorer controller. So it is decoupled a bit more.

I'm hoping that once we file-map on demand that the number of places passed the file mapper goes away.

Trying to take the decoupling one step at a time.

---
In reply to: [384333595](https://github.com/microsoft/sarif-vscode-extension/pull/214#issuecomment-384333595) [](ancestors = 384333595)

src/CodeFlowDecorations.ts
```diff
@@ -59,21 +62,20 @@ export class CodeFlowDecorations implements Disposable {
      * Updates the GutterIcon for the current active Diagnostic
      */
     public updateResultGutterIcon(): void {
```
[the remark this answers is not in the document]
GabeDeBacker: I'm not familiar with that pattern. Can you point me to some documentation on it?

---
In reply to: [400581228](https://github.com/microsoft/sarif-vscode-extension/pull/214#discussion_r400581228) [](ancestors = 400581228)

GabeDeBacker:
And, yes, eventually (probably in the next PR) the FileMapper WILL likely become a singleton object (again) but it's "static" instance will be self-contained and created on demand. (Meaning yep, static again, but in a cleaner way).

I am trying to make this PRs not super huge.

---
In reply to: [606834600](https://github.com/microsoft/sarif-vscode-extension/pull/214#issuecomment-606834600) [](ancestors = 606834600,384333595)

jeffersonking:
It seems it went back and forth (proper singletons, pseudo singletons, back to proper). I'm not quite following the "why" of this. It might be more interesting for me to just look and the next PR and compare it against pre-PR211.

src/CodeFlowDecorations.ts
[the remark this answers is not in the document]
jeffersonking: [Destructuring assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment). It's a less repetitive equivalent of `const activeDiagnostic = this. activeDiagnostic`.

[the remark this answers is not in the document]
GabeDeBacker: I am so confused. This is an assignment into a member variable, not destructing "this". What am I missing?

---
In reply to: [401210654](https://github.com/microsoft/sarif-vscode-extension/pull/214#discussion_r401210654) [](ancestors = 401210654)

PR #216, opened 2020-04-01
Use the NPM package for the SARIF SDK and fix an issue with file-mapping

GabeDeBacker:
Updates the VSCode extension to use the NPM package for the Sarif-MultiTool and removes all the binaries checked in under "resources".

The mult-tool is used for converting from older versions of the schema during load.

With this change, SARIF files created by the Microsoft compiler no longer cause conversion failures (as there was a crash in the multi-tool).

This also fixes an issue where you could never complete file mapping if you gave an absolute path to a file (directories worked), and it now also automatically dismisses the file-mapping UI when a mapping is complete.


src/FileConverter.ts
```diff
@@ -266,11 +265,10 @@ export class FileConverter {
...
 
-        if (!FileConverter.multiTool) {
-            FileConverter.multiTool = FileConverter.extensionContext.asAbsolutePath("/resources/sarif.multitool/Sarif.Multitool.exe");
-        }
-
-        return FileConverter.multiTool;
+        // I'd love to use the type-script "usage" here, but that causes a type-script error
+        // https://www.npmjs.com/package/@microsoft/sarif-multitool
+        // so, I'm just doing what the node package does internally (yuck).
+        return  FileConverter.extensionContext.asAbsolutePath(`node_modules/@Microsoft/sarif-multitool-${process.platform}/Sarif.Multitool.exe`);
```
jeffersonking, line 271: Per the readme, we should be using the default export (which is a string representing the path) instead of hardcoding the path. The path may not be guaranteed in the future.

I realize you had issues with `esModuleInterop` but we should try to resolve those properly.

GabeDeBacker: Yep. Done.

Issue #226, opened 2020-04-09
When SARIF conversion fails to upgrade schemas, all of the error details are not collected


GabeDeBacker:
When the SARIF view fails to "upgrade" a SARIF file that has an earlier schema, it does not properly collect all the data written to standard out and only shows the error code.

The user is left with "Sarif upgrade failed with error:" and nothing else.

PR #227, opened 2020-04-09
Collect standard error while running the conversion tool.

GabeDeBacker:
Quick fix to address issue #226.

When running the conversion tool, the viewer was not collecting all of the information from "standard error". As such, the user would see "Sarif upgrade failed with error:" with nothing else.

### DIFF
--- a/src/FileConverter.ts
+++ b/src/FileConverter.ts
@@ -199,9 +199,13 @@ export class FileConverter {
         const fileOutputPath: string = output;
         const errorData: string[] = [];
         const converted: boolean = await new Promise<boolean>((resolve) => {
-            const proc: ChildProcess = spawn(multiToolPath,
-                ["transform", doc.uri.fsPath, "-o", fileOutputPath, "-p", "-f"],
-            );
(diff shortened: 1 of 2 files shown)


src/FileConverter.ts
```diff
@@ -199,9 +199,13 @@ export class FileConverter {
         const fileOutputPath: string = output;
         const errorData: string[] = [];
         const converted: boolean = await new Promise<boolean>((resolve) => {
-            const proc: ChildProcess = spawn(multiToolPath,
-                ["transform", doc.uri.fsPath, "-o", fileOutputPath, "-p", "-f"],
-            );
+            const launchMultiToolPath: string = multiToolPath;
```
jeffersonking, line 202: Why this practice of introducing an explaining variable for another variable (that seems to me at least) is sufficiently explained?

GabeDeBacker: I can revert this... I was trying to debug the issue and wanted to see the individual steps. No reason for this change.

GabeDeBacker: Reverted

Issue #225, opened 2020-04-09
Results pane does not always correctly select result when selection moves in editor


GabeDeBacker:
Using a sample SARIF file with a large number of results in it, the results pane does not always move to the correct diagnostic when selection changes in the editor.

In some cases, it doesn't change at all.

PR #228, opened 2020-04-09
Remove system JS, fix issues with importing large files, and select the proper diagnostic during code navigation

GabeDeBacker:
This address issue #225, where the result-pane cannot scroll and as the user moves through the source code document the selection does not always cause the correct diagnostic to be selected. This was due to the fact that the "location ranges" in SARIF are largey "empty" (or a single point). So if you manage to put the cursor "just right" in VSCode then it would select. This has been addressed by seeing if the range is "empty" and "single line" and if it is, then just match on the line.

The issue with the result-pane not scrolling (and being empty some times) was due to code that was accidentally left in when trying to find a java-script module loader that would work in vscode's WebView. For note: RequireJS works fine, SystemJS, not so much :). I forget to remove some of the SystemJS code. That was PR #211 

Also, for future referece, when you call spawn in node, it automatically quotes parameters. (Ugh). I thought that was an issue that was causing the multi-tool converter to fail, which I later found is not the cause, it is due to missing .Net dependencies (ugh).

### DIFF
--- a/src/FileConverter.ts
+++ b/src/FileConverter.ts
@@ -200,7 +200,10 @@ export class FileConverter {
         const fileOutputPath: string = output;
         const errorData: string[] = [];
         const converted: boolean = await new Promise<boolean>((resolve) => {
-            const proc: ChildProcess = spawn(multiToolPath, ["transform", `"${doc.uri.fsPath}"`, "-o", `"${fileOutputPath}"`, "-p", "-f"]);
+            // If you are tempted to put quotes around these strings, please don't as "spawn" does that internally.
+            // Something to consider is adding an option to the SARIF viewr so the path to the multi-tool
+            // can be over-ridden for testing.
+            const proc: ChildProcess =  spawn(multiToolPath, ["transform", doc.uri.fsPath, "-o", fileOutputPath, "-p", "-f"]);
 
             proc.stderr.on("data", (data) => {
                 errorData.push(data.toString());
--- a/src/SVDiagnosticCollection.ts
+++ b/src/SVDiagnosticCollection.ts
@@ -340,6 +340,10 @@ export class SVDiagnosticCollection implements Disposable {
         for (const [key, unmappedDiagnostics] of this.unmappedIssuesCollection.entries()) {
             const indexOfUnmappedDiagnostic: number = unmappedDiagnostics.indexOf(diagnostic);
             if (indexOfUnmappedDiagnostic >= 0) {
+                // If we have a large number of diagnostics, deleting the key
+                // and resetting it is far better than having two allocations going at once.
+                this.unmappedIssuesCollection.delete(key);
+
                 unmappedDiagnostics.splice(indexOfUnmappedDiagnostic, 1);
                 this.unmappedIssuesCollection.set(key, unmappedDiagnostics);
 
@@ -347,7 +351,9 @@ export class SVDiagnosticCollection implements Disposable {
                 this.addToCollection(this.mappedIssuesCollection, diagnostic);
 
                 // Attempt to automatically re-map the renamining unmapped diagnostics.
-                remainingUnmappedDiagnostics = unmappedDiagnostics;
+                if (!this.remappingDiagnostics) {
+                    remainingUnmappedDiagnostics = unmappedDiagnostics;
+                }
                 break;
             }
         }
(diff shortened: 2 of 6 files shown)


src/FileConverter.ts
```diff
@@ -200,7 +200,10 @@ export class FileConverter {
         const fileOutputPath: string = output;
         const errorData: string[] = [];
         const converted: boolean = await new Promise<boolean>((resolve) => {
-            const proc: ChildProcess = spawn(multiToolPath, ["transform", `"${doc.uri.fsPath}"`, "-o", `"${fileOutputPath}"`, "-p", "-f"]);
+            // If you are tempted to put quotes around these strings, please don't as "spawn" does that internally.
```
jeffersonking, line 203: Sow how did you end up dealing with spaces in your path?

src/FileConverter.ts
```diff
@@ -200,7 +200,10 @@ export class FileConverter {
...
+            // Something to consider is adding an option to the SARIF viewr so the path to the multi-tool
+            // can be over-ridden for testing.
```
jeffersonking, line 205: Should be able to intercept and mock `imports` during testing. At least it's possible in Jest.

src/SVDiagnosticCollection.ts
```diff
@@ -393,9 +405,15 @@ export class SVDiagnosticCollection implements Disposable {
...
         for (const diagnosticForDocument of diagnosticsForDocument) {
-            if (diagnosticForDocument.location.range.intersection(firstRange)) {
+            // So for single point ranges in SARIF locations (which is apparently common),
+            // If the location range is empty (which means single point) and a single line
+            // and it is on the same line as the selection, then let's use it.
+            // Otherwise, if they intersect, use it as well.
+            if ((diagnosticForDocument.location.range.isEmpty &&
+                diagnosticForDocument.location.range.isSingleLine &&
+                diagnosticForDocument.location.range.start.line === firstRange.start.line) ||
+                (diagnosticForDocument.location.range.intersection(firstRange))) {
```
jeffersonking, line 416: Refactor out `diagnosticForDocument.location`?

src/SVDiagnosticCollection.ts
```diff
@@ -373,14 +379,20 @@ export class SVDiagnosticCollection implements Disposable {
      * that as the active selection which causes pretty much all the UI to update.
      * @param textEditorSelectionChanged The type of text editor selection change.
      */
-    private onDidChangeTextEditorSelection(textEditorSelectionChanged: TextEditorSelectionChangeEvent ): void {
+    private async onDidChangeTextEditorSelection(textEditorSelectionChanged: TextEditorSelectionChangeEvent ): Promise<void> {
```
jeffersonking, line 382: Didn't see any `await`s added.

src/SVDiagnosticCollection.ts
```diff
@@ -340,14 +340,20 @@ export class SVDiagnosticCollection implements Disposable {
         for (const [key, unmappedDiagnostics] of this.unmappedIssuesCollection.entries()) {
             const indexOfUnmappedDiagnostic: number = unmappedDiagnostics.indexOf(diagnostic);
             if (indexOfUnmappedDiagnostic >= 0) {
+                // If we have a large number of diagnostics, deleting the key
+                // and resetting it is far better than having two allocations going at once.
```
jeffersonking, line 344: What would be the "two allocations going at once"?

jeffersonking:
Regarding module loading, what issues were you having (in Rusty's version) before you switched to RequireJS?

src/FileConverter.ts
GabeDeBacker: Right, but the goal is to be able to test the "real Marketplace extension" with work in the SDK in progress. I'll open an issue with comments. Certainly not doing it in this PR.

---
In reply to: [406538342](https://github.com/microsoft/sarif-vscode-extension/pull/228#discussion_r406538342) [](ancestors = 406538342)

src/SVDiagnosticCollection.ts
GabeDeBacker: Yeah, there isn't one. I've removed this comment and the delete.

Also, this code had an issue with the way it was remapping. You cannot "splice" and iterate async at the same time (that results in strange behavior obviously).

We now keep track of "what" needs to be moved between the collections and do that after automatic remapping has completed.

---
In reply to: [406540906](https://github.com/microsoft/sarif-vscode-extension/pull/228#discussion_r406540906) [](ancestors = 406540906)

GabeDeBacker:
When adding more imports for typing in typescript, that causes the generated code to require a different module loader. The module loader required depends on how you configure tsconfig.explorer.json.

We can certainly revisit this for sure. All I'm after is making sure the web-view code maintains type-safety.

---
In reply to: [611812837](https://github.com/microsoft/sarif-vscode-extension/pull/228#issuecomment-611812837) [](ancestors = 611812837)

src/FileConverter.ts
jeffersonking: Understood. That was definitely not my immediate interpretation of the comment phrasing.

Issue #234, opened 2020-04-10
SARIF view code needs to be at least "set up" for localization.


GabeDeBacker:
Currently there are hard-coded strings everywhere in the SARIF view code.

It should consider using vscode-nls-dev framework to at least be "able" to be localized. The earlier we do this, the better as it will become harder and harder as more code is added.

GabeDeBacker:
A simple example of doing this can be found here:
https://github.com/microsoft/vscode-extension-samples/tree/master/i18n-sample

PR #235, opened 2020-04-11
Prepare the extension for localization

GabeDeBacker:
At some point this extension should likely be localized to other languages. This address #234 

There were hard-coded user facing strings, well, all over the place :)

This change moves to using VSCode's 'vscode-nls' package (and it's nls.localize calls) so that it can be used later with another package (vscode-nls-dev) to perform localization.

It also enables TSLint rules to help catch developers typing in user-facing strings.

The way that works is simple. If it detects a string with a double quotation mark, it assumes it will be user facing and warns. If it is a string with a single quotation mark, it assumes it is okay to be a hard-coded (non-user facing) string. Now, of course, if you type a user-facing string in single quotes, it won't catch those, so code-reviews are still important! (And yep, we had some hiding out like that).

The user-facing strings are then "wrapped" in a call like this:
```typescript
localize("someMessageKey", "My user facing text that says {0}", someString)
````

Then "someMessageKey" is placed in package.nls.json as well (Exactly why this is necessary I am tracking down because I'm curious):
```json
"someMessageKey": "My user facing text that says {0}"
```

What was surprising to me with this change is that I had to enable some Microsoft extension to TSLint (and yes, we are using the vscode-tslint :)) to get the warning about needing to localize strings. I then had to proceed to disable, well, a ton of warnings because I wanted to limit this pull-request to just the localization changes.

@jeffersonking Important note: This does not change any of the code in the web-view and not surprisingly there are lot of issues there.


src/factories/locationFactory.ts
```diff
@@ -98,8 +98,8 @@ export namespace LocationFactory {
      * @param resultIndex the index of the result in the SARIF file
      */
     export function mapToSarifFileResult(sarifJSONMapping: Map<string, JsonMapping>, sarifUri: Uri, runIndex: number, resultIndex: number): Location {
-        const resultPath: string = "/runs/" + runIndex + "/results/" + resultIndex;
-        return LocationFactory.createLocationOfMapping(sarifJSONMapping, sarifUri, resultPath, true);
+        const resultPath: string = `'/runs/${runIndex}/results/${resultIndex}`;
```
GabeDeBacker, line 101: >' [](start = 36, length = 1)

This needs to be removed.

tslint.json
```diff
@@ -40,7 +45,12 @@
                 "variable-declaration": "nospace"
             }
         ],
+        "linebreak-style": {"options": ["CRLF"]},
```
jeffersonking, line 48: Make the Mac guy go CRLF huh? :-)

GabeDeBacker: Lol. The files were already CRLF, so... :) This was picked up from our internal extension's settings.

---
In reply to: [407448772](https://github.com/microsoft/sarif-vscode-extension/pull/235#discussion_r407448772) [](ancestors = 407448772)

package.nls.json
```diff
@@ -0,0 +1,124 @@
...
+    "openRemappingInputDialog.skipTooltip": "Skip to next",
+    "openRemappingInputDialog.openDialogLabel": "Map",
+    "openRemappingInputDialog.alreadyExistsInRoot": "'{0}' already exists in the settings (sarif-viewer.rootpaths), please try a different path (Press 'Escape' to cancel)",
+
+    "logReader.proecssingTitle": "Processing {0}",
+    "logReader.processingSarifFile": "Parsing Sarif file",
+    "logReader.jsonFileReadingError": "Sarif Viewer: Cannot display results for '{0}' because: {1}",
+    "logReader.schemaNotDefined": "Sarif Viewer: Cannot display results for '{0}' because the shema was not defined.",
+    "logReader.loadingResults": "Loading {0} Results",
+    "logReader.loadingResultsPercentage": "`Loading {0} Results: {1}% completed`",
```
GabeDeBacker, line 67: >` [](start = 43, length = 1)

Fix this

src/codeFlowDecorations.ts
```diff
@@ -346,38 +348,38 @@ export class CodeFlowDecorations implements Disposable {
...
-        overviewRulerColor: "red",
+        overviewRulerColor: `red`,
         overviewRulerLane: OverviewRulerLane.Left,
         rangeBehavior: DecorationRangeBehavior.ClosedClosed,
     });
 
     private static UnimportantLocationDecorationType = window.createTextEditorDecorationType({
         dark: {
-            backgroundColor: "rgba(150,150,150,.4)",
+            backgroundColor: `rgba(150,150,150,.4)`,
```
jeffersonking, line 377: Single quotes?

tslint.json
GabeDeBacker, line 48: Remove this.

src/codeFlowDecorations.ts
GabeDeBacker: Yes. Fixed, everywhere

---
In reply to: [407679828](https://github.com/microsoft/sarif-vscode-extension/pull/235#discussion_r407679828) [](ancestors = 407679828)

Issue #230, opened 2020-04-10
Consider providing ability to override Multi-Tool conversion tool path


GabeDeBacker:
It is possible to work issue #229, by cloning the SARIF-SDK repository and build the multi-tool yourself.

However, it requires code change (or mocking) to launch the multi-tool from a different location.

Adding the ability to override the multi-tool path would allow testing of a new SDK with the Plugin in the market-place as well as ease development.

PR #236, opened 2020-04-11
Add a simple way for the SARIF-SDK multi-tool path to be overridden

GabeDeBacker:
This address issue #230. Currently the multi-tool converter is not working correctly on certain machines when used from the NPM package. @michaelcfanning is looking into that.

In order to unblock, well, myself and potentially others, this PR adds the ability to override the path to the "mulit-tool" in the process environment block. This will allow testing of changes of the SDK with versions of the extensions that have already been published to the market place as well.

I considered a user option, but, we probably don't want to easily expose this as a user setting.

### DIFF
--- a/src/FileConverter.ts
+++ b/src/FileConverter.ts
@@ -201,9 +201,7 @@ export class FileConverter {
         const errorData: string[] = [];
         const converted: boolean = await new Promise<boolean>((resolve) => {
             // If you are tempted to put quotes around these strings, please don't as "spawn" does that internally.
-            // Something to consider is adding an option to the SARIF viewr so the path to the multi-tool
-            // can be over-ridden for testing.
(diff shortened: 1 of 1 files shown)


src/FileConverter.ts
```diff
@@ -380,4 +378,13 @@ export class FileConverter {
 
         return FileConverter.parseVersion(rawSchemaVersion);
     }
+
+    /**
+     * Returns the path to the multi-tool.
+     * The path can be overridden by setting 'sarifViewer.multiToolPath'
+     * in the process environment.
+     */
+    private static get pathToMulitTool(): string {
```
jeffersonking, line 387: sp

jeffersonking: Why `pathToMultiTool` instead of `multiToolPath`? There's no naming collision.

GabeDeBacker: This was due to the "multiToolPath" "default import" above.

I've aliased the default import to "nodeModulesMultiToolPath" and change this property name to "multiToolPath" to make it very clear what this doing.

Issue #233, opened 2020-04-10
Selecting a new issue\diagnostic does not scroll it into view in the "result" explorer


GabeDeBacker:
When a navigating through the source code (which will set an active diagnostic), or by selecting one in VSCode's problems pane, then active diagnostic is not scrolled into view in the "Result" explorer.

PR #237, opened 2020-04-11
Scroll the selected result into view and expand its group

GabeDeBacker:
This address issue #233 
While testing the extension with a large number of results, it become readily apparent that when the user is navigating through the source code (which causes results to be selected in the viewer), the viewer did not expand or scroll the result into view (even though it was selected).

This is a quick fix to address that while @jeffersonking is working on improvements to the viewer.

### DIFF
--- a/src/FileConverter.ts
+++ b/src/FileConverter.ts
@@ -203,7 +203,7 @@ export class FileConverter {
             // If you are tempted to put quotes around these strings, please don't as "spawn" does that internally.
             // Something to consider is adding an option to the SARIF viewr so the path to the multi-tool
             // can be over-ridden for testing.
-            const proc: ChildProcess =  spawn(multiToolPath, ["transform", doc.uri.fsPath, "-o", fileOutputPath, "-p", "-f"]);
+            const proc: ChildProcess =  spawn("d:\\Gabriel\\Documents\\Repos\\sarif-sdk\\bld\\bin\\AnyCPU_Debug\\Sarif.Multitool\\net461\\Sarif.Multitool.exe", ["transform", doc.uri.fsPath, "-o", fileOutputPath, "-p", "-f"]);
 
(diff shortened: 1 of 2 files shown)


src/explorer/resultsList.ts
```diff
@@ -621,6 +624,31 @@ export class ResultsList {
...
+     * Toggles the group row as well as any result rows that match the group
+     * @param row Group row to toggled
+     */
+    private expandGroupForResult(row: HTMLTableRowElement): void {
+
+        if (row.dataset.group === undefined) {
+            throw new Error("Expected to have collapse group.");
+        }
+
+        if (row.classList.contains(`${ToggleState.collapsed}`)) {
```
jeffersonking, line 637: Don't think the enum->string coercion is necessary.

GabeDeBacker: Agreed. Fixed here and elsewhere.

src/FileConverter.ts
```diff
@@ -203,7 +203,7 @@ export class FileConverter {
             // If you are tempted to put quotes around these strings, please don't as "spawn" does that internally.
             // Something to consider is adding an option to the SARIF viewr so the path to the multi-tool
             // can be over-ridden for testing.
-            const proc: ChildProcess =  spawn(multiToolPath, ["transform", doc.uri.fsPath, "-o", fileOutputPath, "-p", "-f"]);
+            const proc: ChildProcess =  spawn("d:\\Gabriel\\Documents\\Repos\\sarif-sdk\\bld\\bin\\AnyCPU_Debug\\Sarif.Multitool\\net461\\Sarif.Multitool.exe", ["transform", doc.uri.fsPath, "-o", fileOutputPath, "-p", "-f"]);
```
roywill, line 206: >pawn("d:\\Gabriel\\Document [](start = 41, length = 27)

Oops

PR #239, opened 2020-04-13
Do not alias the default import

GabeDeBacker:
When completing pull request #236 , I attempted to alias the default import with a different name so that the static property in the FileConverter did not have the same as the default import.

Unfortunately while testing, I still had the environment variable set so I wasn't actually using the alias.

While working with @michaelcfanning to figure out why the multi-tool isn't working from the NPM package on my machine, I unset the environment variable and realized the alias wasn't working as expected. Instead of returning a string, it was returning an object where all the path characters where individual characters rather than a string???? (I'll leave it to the javascript module exports to explain that one).

This change reverts the alias.

### DIFF
--- a/package.json
+++ b/package.json
@@ -207,10 +207,10 @@
         "tslint-no-unused-expression-chai": "^0.1.4"
     },
     "dependencies": {
+        "@microsoft/sarif-multitool": "^2.2.2",
         "json-source-map": "^0.4.0",
         "jsonpointer": "4.0.1",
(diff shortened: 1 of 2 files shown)


jeffersonking:
The issue you describe "individual characters rather than a string" is commonly caused by:

`import * as multitoolPath from '@microsoft/sarif-multitool'`

as opposed to:

`import multitoolPath from '@microsoft/sarif-multitool'`

Issue #242, opened 2020-04-14
Clean up remaining lint issues


GabeDeBacker:
After turning on more linting for the extension code during this pull request #235 , there are a number of lint issues remaining. We should clean those up

PR #243, opened 2020-04-14
Turn on strict type predicate

GabeDeBacker:
Starts to address issue #242. Turns on "strict type predicate" which performs linting operations such as verifying that "if" conditions for "undefined" or against "types" do not evaluate to constant true or false expressions.

### DIFF
--- a/src/fileMapper.ts
+++ b/src/fileMapper.ts
@@ -137,7 +137,7 @@ export class FileMapper implements Disposable {
             this.addToFileMapping(Utilities.getFsPathWithFragment(origUri), uri);
             this.saveBasePath(origUri, uri, uriBase);
             this.fileRemapping.forEach((value: Uri | undefined, key: string) => {
-                if (value === null) {
+                if (value) {
                     this.tryRebaseUri(Uri.file(key));
                 }
             });
--- a/src/fileMapper.ts
+++ b/src/fileMapper.ts
@@ -137,7 +137,7 @@ export class FileMapper implements Disposable {
             this.addToFileMapping(Utilities.getFsPathWithFragment(origUri), uri);
             this.saveBasePath(origUri, uri, uriBase);
             this.fileRemapping.forEach((value: Uri | undefined, key: string) => {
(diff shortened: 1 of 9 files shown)


src/fileMapper.ts
```diff
@@ -137,7 +137,7 @@ export class FileMapper implements Disposable {
             this.addToFileMapping(Utilities.getFsPathWithFragment(origUri), uri);
             this.saveBasePath(origUri, uri, uriBase);
             this.fileRemapping.forEach((value: Uri | undefined, key: string) => {
-                if (value === null) {
+                if (value) {
```
roywill, line 140: Inverted?

roywill:
>             if (!this.groupBy) {

Seems like a waste of time.

---
Refers to: src/resultsListController.ts:439 in 4419c9c. [](commit_id = 4419c9ca92004ac3500f4f9e15e9a046108edbe3, deletion_comment = False)

src/resultsListController.ts
```diff
@@ -440,7 +440,11 @@ export class ResultsListController implements Disposable {
                 continue;
             }
```
GabeDeBacker, line 441: This condition is always false as well

src/resultsListController.ts
```diff
@@ -480,26 +484,29 @@ export class ResultsListController implements Disposable {
                     valueB = a[this.sortBy.column];
                 }
 
-                if (valueA === undefined || valueA.value === undefined) {
-                    comp = -1;
-                    if (valueB === undefined || valueB.value === undefined) {
+                if (valueA === undefined ||
```
roywill, line 487: Pull this into an equality function.

GabeDeBacker: Doing so would remove the TypeScript compiler's ability to know they are not undefined on line 498.

I've made this code easier to understand though.

---
In reply to: [408282753](https://github.com/microsoft/sarif-vscode-extension/pull/243#discussion_r408282753) [](ancestors = 408282753)

Issue #244, opened 2020-04-14
Add command to allow other extensions to open SARIF files


GabeDeBacker:
There are external extensions that produce SARIF files and currently the only way to view the SARIF results is to open the SARIF file in the editor.

An example would be an extension that performs it's own static analysis during a "build".

The concept here would be to allow an extension to direct the SARIF extension to parse and open the SARIF results in the viewer.

The arguments to the command would include:

- Specifying multiple URIs to SARIF files
- Disabling the schema upgrade prompt (the files would upgraded automatically)
- Allowing the caller to configure whether the SARIF document is opened in an editor
- An option to not open the SARIF viewer if there are no results in the parsed SARIF file.

GabeDeBacker:
This issue will remain open until we decide if we want to publish an NPM\Node package for the API

Issue #229, opened 2020-04-10
The multi-tool converter in the new SARIF SDK npm package fails


GabeDeBacker:
SARIF files that need to be upgraded to the latest schema version fail when the “multi-tool” is run from the NPM package, this exception occurs:
Error:
  An assembly specified in the application dependencies manifest (Sarif.Multitool.deps.json) was not found:
    package: 'CsvHelper', version: '15.0.0'
    path: 'lib/netstandard2.1/CsvHelper.dll'

PR #248, opened 2020-04-15
Expose API for other plugins to drive the SARIF viewer

GabeDeBacker:
This change exposes an API for other extension to pass "SARIF" files to the extension to process and load into the viewer. It address isssue #244 and issues #229 .

The API should eventually be turned into a "typings" NODE package and published so it can be downloaded through NPM as "microsoft/sarf-viewer-api".

I want to hold off on that until we proove out not only the API with our dependent extension owners, but also verify that we like the names, function signatures, etc.

This has been verified to function end to end with another extension developed at Microsoft.

### DIFF
--- a/package.json
+++ b/package.json
@@ -195,6 +195,14 @@
                 "command": "extension.sarif.previousCodeFlowStep",
                 "title": "%commands.extension.sarif.previousCodeFlowStep.title%",
                 "category": "%commands.category%"
+            },
+            {
+                "command": "extension.sarif.openLogFiles",
+                "title": "%commands.extension.sarif.openeLogFiles"
+            },
+            {
+                "command": "extension.sarif.closeLogFiles",
+                "title": "%commands.extension.sarif.closeLogFiles"
             }
         ],
         "keybindings": [
--- a/package.json
+++ b/package.json
@@ -195,14 +195,6 @@
                 "command": "extension.sarif.previousCodeFlowStep",
                 "title": "%commands.extension.sarif.previousCodeFlowStep.title%",
                 "category": "%commands.category%"
-            },
-            {
-                "command": "extension.sarif.openLogFiles",
-                "title": "%commands.extension.sarif.openeLogFiles"
-            },
-            {
-                "command": "extension.sarif.closeLogFiles",
-                "title": "%commands.extension.sarif.closeLogFiles"
             }
         ],
         "keybindings": [
--- a/package.json
+++ b/package.json
@@ -249,7 +249,7 @@
         "tslint": "^6.1.1"
     },
     "dependencies": {
-        "@microsoft/sarif-multitool": "^2.2.2",
+        "@microsoft/sarif-multitool": "^2.2.3",
         "json-source-map": "^0.4.0",
         "jsonpointer": "4.0.1",
         "markdown-it": "^9.0.1",
--- a/src/api/sarifViewerApi.ts
+++ b/src/api/sarifViewerApi.ts
@@ -1,0 +1,56 @@
+/*!
+ * Copyright (c) Microsoft Corporation. All Rights Reserved.
+ */
+
+import * as vscode from 'vscode';
+
+/**
+ * The expected arguments for the @see openLogFileCommand.
+ */
+export interface OpenLogFileArguments {
+    /**
+     * The array of SARIF log files to open. The log file URIs must be of scheme 'file' and be valid SARIF JSON files.
+     */
+    readonly sarifFiles: vscode.Uri[];
+
+    /**
+     * Specifies options indicating whether to prompt the user when the
+     * SARIF log file shcmea is needs to be ugpraded to the latest one.
+     * If not specified, the user will be prompted. (Default is true).
+     */
+    readonly promptUserForUpgrade?: boolean;
+
+    /**
+     * Indicates whether to open the files (if not already open) and display it to the user.
+     * If not specified, the files will be displayed. (Default is true).
+     */
+    readonly openInTextEditor?: boolean;
+
+    /**
+     * Indicates whether the SARIF viewer should upen when there are no results to display.
+     * If not specificm, the viewer will be displayed. (Default is true).
+     */
+    readonly openViewerOnNoResults?: boolean;
+}
+
+/**
+ * The command the SARIF extension exposes to open SARIF log files.
+ * The expected arguments are of type @see OpenLogFileArguments
+ */
+export const openLogFilesCommand: string = 'extension.sarif.openLogFiles';
+
+/**
+ * The expected arguments for the @see closeLogFileCommand.
+ */
+export interface CloseLogFileArguments {
+    /**
+     * The array of SARIF log files to close.
+     */
+    readonly sarifFiles: vscode.Uri[];
+}
+
+/**
+ * The command the SARIF extension exposes to open SARIF log files.
+ * The expected arguments are of type @see CloseLogFileArguments
+ */
+export const closeLogFilesCommand: string = 'extension.sarif.closeLogFiles';
--- a/src/api/sarifViewerApi.ts
+++ b/src/api/sarifViewerApi.ts
@@ -13,13 +13,13 @@ export interface OpenLogArguments {
      * SARIF log schema is needs to be upgraded to the latest one schema version.
      * The default is true.
      */
-    readonly promptUserForUpgrade?: boolean;
+    readonly promptForUpgrade?: boolean;
 
     /**
      * Indicates whether the SARIF viewer should open when there are no results in the logs to display.
      * The default is true.
      */
-    readonly openViewerOnNoResults?: boolean;
+    readonly openViewerWhenNoResults?: boolean;
 }
 
 /**
@@ -27,20 +27,20 @@ export interface OpenLogArguments {
  */
 export interface Api {
     /**
-     * Instructs the SARIF viewer to open logs.
+     * Opens logs .
      * @param logs The logs to open.
      * @param openLogArguments Parameters that control how the logs are opened.
      */
     openLogs(logs: vscode.Uri[], openLogArguments?: OpenLogArguments): Promise<void>;
 
     /**
-     * Instructs the SARIF viewer to close logs.
+     * Closes logs.
      * @param logs The log to close.
      */
     closeLogs(logs: vscode.Uri[]): Promise<void>;
 
     /**
-     * Instructs the SARIF viewer to close all logs.
+     * Closes all logs.
      */
     closeAllLogs(): Promise<void>;
 }
--- a/src/api/sarifViewerApi.ts
+++ b/src/api/sarifViewerApi.ts
@@ -4,9 +4,6 @@
 
 import * as vscode from 'vscode';
 
-/**
- * The expected arguments for the @see openLogFileCommand.
- */
 export interface OpenLogFileArguments {
     /**
      * The array of SARIF log files to open. The log file URIs must be of scheme 'file' and be valid SARIF JSON files.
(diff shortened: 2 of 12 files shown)


package.json
```diff
@@ -241,7 +241,7 @@
         "tslint": "^6.1.1"
     },
     "dependencies": {
-        "@microsoft/sarif-multitool": "^2.2.2",
+        "@microsoft/sarif-multitool": "^2.2.3",
```
GabeDeBacker, line 244: >    "@microsoft/sarif-multitool": "^2.2.3", [](start = 4, length = 43)

This fixes the upgrade tool not converting. Not strictly part of this PR, but it makes it work end to end.

src/test/locationFactory.test.ts
```diff
@@ -10,7 +10,7 @@ import { LocationFactory } from "../factories/locationFactory";
 
 suite("parseRange", () => {
     test("Undefined range", async () => {
-        const region: sarif.Region | undefined = undefined;
+        const region: sarif.Region = {};
```
GabeDeBacker, line 13: Parse range does not take undefined. TSLint is disabled for test code, which is why this wasn't caught. We should consider enabling tslint for test code.

src/api/sarifViewerApi.ts
```diff
@@ -0,0 +1,46 @@
...
+     * SARIF log file shcmea is needs to be ugpraded to the latest one.
+     * If not specified, the user will be prompted. (Default is true).
+     */
+    readonly promptUserForUpgrade?: boolean;
+
+    /**
+     * Indicates whether to open the files (if not already open) and display it to the user.
+     * If not specified, the files will be displayed. (Default is true).
+     */
+    readonly openInTextEditor?: boolean;
```
jeffersonking, line 19: "Files" means Artifacts, right?

I assume "File" does not mean SARIF Log. But then again we are inside `OpenLogFileArguments`.

Could we disambiguate?

src/utilities/apiImpl.ts
```diff
@@ -0,0 +1,51 @@
...
+
+    /**
+     * @inheritdoc
+     */
+    public async openLogFiles(sarifFiles: vscode.Uri[], openLogFileArguments?: sarifApi.OpenLogFileArguments): Promise<void> {
+        for (const sarifUri of sarifFiles) {
+            await openSarifFile(sarifUri, this.logReader, this.diagnosticCollection, {
+                closeOriginalFileOnUpgrade: true,
+                openInTextEditor: openLogFileArguments?.openInTextEditor === undefined ? true : openLogFileArguments.openInTextEditor,
+                promptUserForUpgrade: openLogFileArguments?.promptUserForUpgrade === undefined ? true : openLogFileArguments.promptUserForUpgrade
```
jeffersonking, line 28: `openLogFileArguments?.promptUserForUpgrade ?? true` #pattern

src/utilities/apiImpl.ts
```diff
@@ -0,0 +1,51 @@
...
+
+        return Promise.resolve();
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public async closeAllLogFiles(): Promise<void> {
+        this.diagnosticCollection.removeAllRuns();
+        return Promise.resolve();
```
jeffersonking, line 49: Is your convention to return `Promise.resolve()` rather than just `undefined` (which the `async` will wrap for you)? #pattern

On a non-void async function, would you likewise return `Promise.resolve(whateverValue)`?

src/api/sarifViewerApi.ts
```diff
@@ -0,0 +1,46 @@
...
+
+import * as vscode from 'vscode';
+
+export interface OpenLogFileArguments {
+    /**
+     * Specifies options indicating whether to prompt the user when the
+     * SARIF log file shcmea is needs to be ugpraded to the latest one.
+     * If not specified, the user will be prompted. (Default is true).
+     */
+    readonly promptUserForUpgrade?: boolean;
```
jeffersonking, line 13: Would hope to be more systematic in the phrasing here. The other two start with `Indicates weather...", yet this one is different for no apparent reason.

Also, a lone "Default is true." seem most clear. IMO redundantly adding "If not specified, the user will be prompted." reduces clarity.

src/extension.ts
```diff
@@ -135,7 +139,7 @@ async function openSarifFileIfNotOpen(sarifFile: vscode.Uri): Promise<void> {
  * @param diagnosticCollection The diagnostic collection to add the results too.
  * @param options Options controlling the upgrade prompt if a SARIF file is at an earlier schema version.
  */
-async function openSarifFile(sarifFile: vscode.Uri,  logReader: LogReader, diagnosticCollection: SVDiagnosticCollection, options: OpenSarifFileOptions): Promise<void> {
+export async function openSarifFile(sarifFile: vscode.Uri, logReader: LogReader, diagnosticCollection: SVDiagnosticCollection, options: OpenSarifFileOptions): Promise<void> {
     if (!sarifFile.isSarifFile()) {
         return;
```
jeffersonking, line 144: Per my other comment, this one isn't a `Promise.resolve()`?

src/extension.ts
```diff
@@ -18,17 +18,19 @@ import { Utilities } from "./utilities";
 import { ResultsListController } from "./resultsListController";
 import { FileMapper } from "./fileMapper";
 import { SVDiagnosticCollection } from "./svDiagnosticCollection";
+import * as sarifViewrApi from "./api/sarifViewerApi";
```
jeffersonking, line 21: `import { Api } from "./api/sarifViewerApi"`

GabeDeBacker:
Yeah, I need to re-enable the spell checker

---
In reply to: [394116124](https://github.com/microsoft/sarif-vscode-extension/pull/248#issuecomment-394116124) [](ancestors = 394116124)

src/utilities/apiImpl.ts
GabeDeBacker: I need to investigate how openLogFileArguments?.promptUserForUpgrade ?? true does when promptUserForUpgrade is undefined vs false

---
In reply to: [409120203](https://github.com/microsoft/sarif-vscode-extension/pull/248#discussion_r409120203) [](ancestors = 409120203)

src/utilities/apiImpl.ts
GabeDeBacker: Ah.. This function doesn't have an await, but I wanted to leave "Wiggle room" in the API  for it to be async.

If you just do public closeLogFiles(): Promise<void> with no async, the linter complains.. Hmm.. let me look into this.

---
In reply to: [409121885](https://github.com/microsoft/sarif-vscode-extension/pull/248#discussion_r409121885) [](ancestors = 409121885)

src/utilities/apiImpl.ts
GabeDeBacker: works as you demonstrated. Changing.

---
In reply to: [409143005](https://github.com/microsoft/sarif-vscode-extension/pull/248#discussion_r409143005) [](ancestors = 409143005,409120203)

src/extension.ts
GabeDeBacker: This is my preference for sure. The other one was because I wanted to make it possible for the API to be async.

I can\should switch the API to thenable, which allows for either I believe.

---
In reply to: [409124640](https://github.com/microsoft/sarif-vscode-extension/pull/248#discussion_r409124640) [](ancestors = 409124640)

src/utilities/apiImpl.ts
jeffersonking: I mean... keep the `async` modifier. But just change the return value.

src/extension.ts
GabeDeBacker: I added some comments.

---
In reply to: [409145629](https://github.com/microsoft/sarif-vscode-extension/pull/248#discussion_r409145629) [](ancestors = 409145629,409124640)

src/api/sarifViewerApi.ts
```diff
@@ -0,0 +1,52 @@
...
+     * SARIF log schema is needs to be upgraded to the latest one schema version.
+     * The default is true.
+     */
+    readonly promptUserForUpgrade?: boolean;
+
+    /**
+     * Indicates whether to open SARIF logs inside the VSCode editor.
+     * The default is true.
+     */
+    readonly openInEditor?: boolean;
```
GabeDeBacker, line 22: After discussion with Jeff, we believe this can be dropped as we don't currently see any case where this would be passed as true.

src/utilities/apiImpl.ts
```diff
@@ -0,0 +1,52 @@
...
+     * @param logReader An instance of the log reader to use for parsing log files.
+     * @param diagnosticCollection The diagnostic collection that holds the results.
+     */
+    public constructor(private readonly logReader: LogReader, private readonly diagnosticCollection: SVDiagnosticCollection) {
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public async openLogs(sarifFiles: vscode.Uri[], openLogFileArguments?: sarifApi.OpenLogArguments): Promise<void> {
```
sherryyshi, line 23: >Logs [](start = 21, length = 4)

nit: would it be worth specifying what type of logs (e.g. `openSarifFiles`)? or is it obvious that these are sarif files?

src/api/sarifViewerApi.ts
```diff
@@ -0,0 +1,46 @@
...
+     * Indicates whether the SARIF viewer should open when there are no results in the logs to display.
+     * The default is true.
+     */
+    readonly openViewerOnNoResults?: boolean;
+}
+
+/**
+ * API exposed by the SARIF viewer extension.
+ */
+export interface Api {
```
sherryyshi, line 28: >Api [](start = 17, length = 3)

should there be some sort of version info embedded in this? How would backwards compatibility be supported in the future?

src/utilities/apiImpl.ts
GabeDeBacker: Nope. you are already using the SARIF apis, so that would be redundant

---
In reply to: [409154113](https://github.com/microsoft/sarif-vscode-extension/pull/248#discussion_r409154113) [](ancestors = 409154113)

jeffersonking: Another nit: param `sarifFiles` should technically be `uris` or `logUris`.

"SARIF file" is a colloquialism.

src/extension.ts
sherryyshi: it returns a "void" since the function is marked as "async" - the return value should not be a Promise, but just the type encapsulated by the Promise.

---
In reply to: [409147124](https://github.com/microsoft/sarif-vscode-extension/pull/248#discussion_r409147124) [](ancestors = 409147124,409145629,409124640)

src/api/sarifViewerApi.ts
GabeDeBacker: There really doesn't need to be.

As long as you "add" new methods or properties you are good to go as far as JavaScript is concerned.

Heck, you can even add new parameters to the end of an API function and you'd still be good. :)

Point is, once someone is consuming your API, you can't ever really change the signatures and they must remain backward compatible.

---
In reply to: [409156312](https://github.com/microsoft/sarif-vscode-extension/pull/248#discussion_r409156312) [](ancestors = 409156312)

src/utilities/apiImpl.ts
```diff
@@ -0,0 +1,52 @@
...
+        for (const sarifUri of sarifFiles) {
```
roywill, line 24: Why fix the singleton and leave the plural files?

src/utilities/apiImpl.ts
jeffersonking: @sherryyshi There is only one type.

Issue #224, opened 2020-04-09
Viewer doesn't deal well with binary files


lgolding:
1. Load the sample SARIF file [EmbbeddedBinaryContent.sarif](https://raw.githubusercontent.com/microsoft/sarif-samples/master/samples/EmbeddedBinaryContent.sarif).

2. Double-click on its single result.

Expected: Well, it's not clear what to expect here, because I don't know VS Code's facilities for displaying binary files.

Actual: It loads the 5-byte contents into a text editor window and displays squares for the bytes -- which I can't blame it for -- but it doesn't highlight anything, even though the result sepcifies a region with `byteOffset` of 2 and `byteLength` of 2.

GabeDeBacker:
@lgolding - We have addressed this in the current Insiders version as well as the new UI @jeffersonking is working on.

PR #252, opened 2020-04-17
Handle artifact content objects using a file-system provider

GabeDeBacker:
This issues address #224 and also handles the fact that embedded content didn't work well at all.

The implementation using VSCode's concept of a file system provider, which allows an extension to register a URI scheme and then implement the semantics of reading and writing from it.

For our usage, the write portion is not implemented, as for much of the provider because our data is read only.

Using this technique not only removes the need of creating temporary files, but it allows us to render things such as the "binary" content as markdown so the user can have a nice representation of it.

Here's a sample:
image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -31,31 +31,31 @@
             }
         },
         "@microsoft/sarif-multitool": {
-            "version": "2.2.2",
-            "resolved": "https://registry.npmjs.org/@microsoft/sarif-multitool/-/sarif-multitool-2.2.2.tgz",
-            "integrity": "sha512-mid4xKQv5QkrbZtH+V0Tlwl3aCneZtAEoggXqY7klwuza1Zn1uI7CkRqq8aDiv+fWR/a6acAbagEogwzfRVY8A==",
+            "version": "2.2.3",
+            "resolved": "https://registry.npmjs.org/@microsoft/sarif-multitool/-/sarif-multitool-2.2.3.tgz",
+            "integrity": "sha512-5CIfoT8iArzFS3V66bPHNstIxdPk23bZGTu6kDsspH6MLeEAZ+riQ7uJQqdIVEBvFkqJKLrEnRWF6rYzEvz6XQ==",
             "requires": {
-                "@microsoft/sarif-multitool-darwin": "^2.2.2",
-                "@microsoft/sarif-multitool-linux": "^2.2.2",
-                "@microsoft/sarif-multitool-win32": "^2.2.2"
+                "@microsoft/sarif-multitool-darwin": "^2.2.3",
+                "@microsoft/sarif-multitool-linux": "^2.2.3",
+                "@microsoft/sarif-multitool-win32": "^2.2.3"
             }
         },
         "@microsoft/sarif-multitool-darwin": {
-            "version": "2.2.2",
-            "resolved": "https://registry.npmjs.org/@microsoft/sarif-multitool-darwin/-/sarif-multitool-darwin-2.2.2.tgz",
-            "integrity": "sha512-TEWg5Cnb45h23Av77KY9+Pz8AtcN+KjyuL08LbrACrIZ+9n79XnP3GEEhg3SSdY/MizwnD40FxPYbbryYKFdYw==",
+            "version": "2.2.3",
+            "resolved": "https://registry.npmjs.org/@microsoft/sarif-multitool-darwin/-/sarif-multitool-darwin-2.2.3.tgz",
+            "integrity": "sha512-6y3zTVX4qO4w+l/SfSkQaWQuPoUeWE4zykKTNTFLCimudjCsRWE0O4bzbXRbBM1zgG/cbqBBQuvmoaxUOS3iVg==",
             "optional": true
         },
         "@microsoft/sarif-multitool-linux": {
-            "version": "2.2.2",
-            "resolved": "https://registry.npmjs.org/@microsoft/sarif-multitool-linux/-/sarif-multitool-linux-2.2.2.tgz",
-            "integrity": "sha512-dHVfpI1pLv8eVe1WG/v7Dah7lM2FYMXbtzGp590EbhsM4ho7doV1PPyGkZu6jQwafl8FIsk3gwS/z6tCx3G+uQ==",
+            "version": "2.2.3",
+            "resolved": "https://registry.npmjs.org/@microsoft/sarif-multitool-linux/-/sarif-multitool-linux-2.2.3.tgz",
+            "integrity": "sha512-2dU5w7sdvJBFAkCJ9lghEShWEJeSDRFkpC1tul4f2yYVpw8mGBUHyfvxrO0hingiUggAfh0Wt/yXW1YVUjKH6w==",
             "optional": true
         },
         "@microsoft/sarif-multitool-win32": {
-            "version": "2.2.2",
-            "resolved": "https://registry.npmjs.org/@microsoft/sarif-multitool-win32/-/sarif-multitool-win32-2.2.2.tgz",
-            "integrity": "sha512-tY6kV6msjxJCApVKpitUSjn26EqiicxWiQ7Gdc2980fTL0GIN2ZtRChtxIhHGn3eU3XVzGJNSQkhBDYBeqDmXQ==",
+            "version": "2.2.3",
+            "resolved": "https://registry.npmjs.org/@microsoft/sarif-multitool-win32/-/sarif-multitool-win32-2.2.3.tgz",
+            "integrity": "sha512-0CColCvCWvZ6IVq9KSsdhf9tyDmTX/q4fZGoErSluYDHtz1RI3KZFkwfHvq65wlK3FLd2kISQY/cm/kLvmkTFg==",
             "optional": true
         },
         "@types/events": {
--- a/src/artifactContentFileSystemProvider.ts
+++ b/src/artifactContentFileSystemProvider.ts
@@ -7,7 +7,7 @@ import * as sarif from "sarif";
 import * as fs from "fs";
 import * as path from "path";
 import { LogReader } from "./logReader";
-import { ArtifactContentRenderer, tryCreateRendererForArtifactContent } from "./artifactContentRenderers/artifactContentRendering";
+import { ArtifactContentRenderer, tryCreateRendererForArtifactContent } from "./artifactContentRenderers/artifactContentRenderer";
 
 /**
  * The purpose of this class is to provide a "file system provider" to artifact objects
@@ -159,30 +159,12 @@ export class ArtifactContentFileSystemProvider implements vscode.FileSystemProvi
      */
     public async stat(uri: vscode.Uri): Promise<vscode.FileStat> {
         const artifactContent: ArtifactInformationFromUri = await ArtifactContentFileSystemProvider.getArtifactInformationFromUri(uri);
-
-        let contentSize: number = 0;
-        if (artifactContent.contents.text) {
-            contentSize = artifactContent.contents.text.length;
-        } else if (artifactContent.contents.binary) {
-            // This isn't exactly correct as we render it as markdown.
-            // We could make it correct by calling render on the binary renderer
-            // and computing the length. Turns out VSCode doesn't really use this
-            // size, so.... this is good enough.
-            contentSize = artifactContent.contents.binary.length;
-        } else if (artifactContent.contents.rendered) {
-            if (artifactContent.contents.rendered.markdown) {
-                contentSize = artifactContent.contents.rendered.markdown.length;
-            } else if (artifactContent.contents.rendered.text) {
-                contentSize = artifactContent.contents.rendered.text.length;
-            }
-        }
-
         const time: number = artifactContent.artifact.lastModifiedTimeUtc !== undefined ? Date.parse(artifactContent.artifact.lastModifiedTimeUtc) : Date.now();
         return {
             type: vscode.FileType.File,
             ctime: time,
             mtime: time,
-            size: contentSize
+            size: 0
         };
     }
 
@@ -268,6 +250,8 @@ export class ArtifactContentFileSystemProvider implements vscode.FileSystemProvi
             return undefined;
         }
 
+        // Modify the URI path to contain an extension that VSCode will use to detect it's file type
+        // which it then uses for rendering.
         let uriPath: string;
         if (!requiredExtension) {
             uriPath = artifactUri.path;
--- a/src/artifactContentFileSystemProvider.ts
+++ b/src/artifactContentFileSystemProvider.ts
@@ -6,8 +6,8 @@ import * as vscode from "vscode";
 import * as sarif from "sarif";
 import * as fs from "fs";
 import * as path from "path";
-import { BinaryContentRenderer } from "./binaryContentRenderer";
 import { LogReader } from "./logReader";
+import { ArtifactContentRenderer, tryCreateRendererForArtifactContent } from "./artifactContentRenderers/artifactContentRendering";
 
 /**
  * The purpose of this class is to provide a "file system provider" to artifact objects
@@ -200,25 +200,10 @@ export class ArtifactContentFileSystemProvider implements vscode.FileSystemProvi
      */
     public async readFile(uri: vscode.Uri): Promise<Uint8Array> {
         const artifactContent: ArtifactInformationFromUri = await ArtifactContentFileSystemProvider.getArtifactInformationFromUri(uri);
-        if (artifactContent.contents.text) {
-            return new Buffer(artifactContent.contents.text, artifactContent.artifact.encoding);
-        }
+        const artifactContentRenderer: ArtifactContentRenderer | undefined = tryCreateRendererForArtifactContent(artifactContent.log, artifactContent.contents, artifactContent.parsedUriData.runIndex, artifactContent.parsedUriData.artifactIndex);
 
-        if (artifactContent.contents.binary) {
-            const binaryContentRenderer: BinaryContentRenderer | undefined = BinaryContentRenderer.tryCreateFromLog(artifactContent.log, artifactContent.parsedUriData.runIndex, artifactContent.parsedUriData.artifactIndex);
-            if (binaryContentRenderer) {
-                return Buffer.from(binaryContentRenderer.renderAsMarkdown(artifactContent.parsedUriData.artifactUri));
-            }
-        }
-
-        if (artifactContent.contents.rendered) {
-            if (artifactContent.contents.rendered.markdown) {
-                return Buffer.from(artifactContent.contents.rendered.markdown, artifactContent.artifact.encoding);
-            }
-
-            if (artifactContent.contents.text) {
-                return Buffer.from(artifactContent.contents.rendered.text, artifactContent.artifact.encoding);
-            }
+        if (artifactContentRenderer) {
+            return Buffer.from(artifactContentRenderer.render(artifactContent.parsedUriData.artifactUri));
         }
 
         throw new Error(`There is no contents that can be rendered associated with artifact index ${artifactContent.parsedUriData.runIndex} for run index ${artifactContent.parsedUriData.runIndex}.`);
@@ -267,8 +252,9 @@ export class ArtifactContentFileSystemProvider implements vscode.FileSystemProvi
      * @param artifactUri The file name that VSCode will display in the editor and use for detection of type.
      * @param runIndex The index of the run in the SARIF file.
      * @param artifactIndex The artifact index.
+     * @param artifactContentRenderer Indicates how to render binary content.
      */
-    public static tryCreateUri(sarifLog: sarif.Log, logPath: vscode.Uri, artifactUri: vscode.Uri, runIndex: number, artifactIndex: number): vscode.Uri | undefined {
+    public static tryCreateUri(sarifLog: sarif.Log, logPath: vscode.Uri, artifactUri: vscode.Uri, runIndex: number, artifactIndex: number, artifactContentRenderer: ArtifactContentRenderer | undefined): vscode.Uri | undefined {
         if (!logPath.isSarifFile()) {
             throw new Error(`${logPath.toString()} is not a SARIF file`);
         }
@@ -282,9 +268,14 @@ export class ArtifactContentFileSystemProvider implements vscode.FileSystemProvi
             return undefined;
         }
 
-        // We render binary and markdown content as markdown, so
-        // we add the ".md" extension for VSCode to detect that.
-        const vscodeUri: vscode.Uri = (artifact.contents.binary || artifact.contents.rendered?.markdown) ? artifactUri.with({ path: artifactUri.path.concat('.md') }) : artifactUri;
+        let uriPath: string;
+        if (!artifactContentRenderer || !artifactContentRenderer.specificUriExtension) {
+            uriPath = artifactUri.path;
+        } else {
+            const vscodeUri: vscode.Uri = artifactUri.with({ path: artifactUri.path.concat(artifactContentRenderer.specificUriExtension) });
+            uriPath = path.posix.join(artifactUri.path, vscodeUri.path);
+        }
+
         const encodedUriData: EncodedUriData = {
             artifactIndex,
             runIndex,
@@ -293,7 +284,6 @@ export class ArtifactContentFileSystemProvider implements vscode.FileSystemProvi
         };
 
         const encodedUriDataAsBase64: string = new Buffer(JSON.stringify(encodedUriData)).toString('base64');
-        const uriPath: string = path.posix.join(artifactUri.path, vscodeUri.path);
 
         return vscode.Uri.parse(`${ArtifactContentFileSystemProvider.ArtifactContentScheme}://${uriPath}?${encodedUriDataAsBase64}`, /*strict*/ true);
     }
--- a/src/artifactContentFileSystemProvider.ts
+++ b/src/artifactContentFileSystemProvider.ts
@@ -10,6 +10,21 @@ import { BinaryContentRenderer } from "./binaryContentRenderer";
 import { LogReader } from "./logReader";
 
 /**
+ * The purpose of this class is to provide a "file system provider" to artifact objects
+ * which have a "contents" property set ultimate results user display.
+ * Starting from the spec
+ * https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html
+ * An artifact can have "contents" property:
+ * https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc34317619
+ * The artifact contents object can have many different types of rendering.
+ * https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc34317422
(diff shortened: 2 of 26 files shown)


jeffersonking:
Haven't really looked yet, but why is the `package-lock.json` changing when you haven't changed the `package.json`?

roywill:
Gabriel are you thinking of allowing visualization similar to list?

---
In reply to: [615401093](https://github.com/microsoft/sarif-vscode-extension/pull/252#issuecomment-615401093) [](ancestors = 615401093)

GabeDeBacker:
This is as virtualized as you can get :) The content isn't read until it is rendered.

As for the package.lock.json... it looks like it never got updated in master when I updated to multi-tool 2.23, honestly not sure why.

---
In reply to: [615403631](https://github.com/microsoft/sarif-vscode-extension/pull/252#issuecomment-615403631) [](ancestors = 615403631,615401093)

package-lock.json
```diff
@@ -31,31 +31,31 @@
             }
         },
         "@microsoft/sarif-multitool": {
-            "version": "2.2.2",
-            "resolved": "https://registry.npmjs.org/@microsoft/sarif-multitool/-/sarif-multitool-2.2.2.tgz",
-            "integrity": "sha512-mid4xKQv5QkrbZtH+V0Tlwl3aCneZtAEoggXqY7klwuza1Zn1uI7CkRqq8aDiv+fWR/a6acAbagEogwzfRVY8A==",
+            "version": "2.2.3",
```
GabeDeBacker, line 34: >          "version": "2.2.3", [](start = 2, length = 29)

This is is what is causing the package-lock update. Package.json is already updated, I am not sure why this didn't get updated when that occurred.

src/binaryContentRenderer.ts
```diff
@@ -0,0 +1,116 @@
...
+        }
+
+        if (!artifact.contents.binary) {
+            return undefined;
+        }
+
+        return new BinaryContentRenderer(artifact.contents.binary);
+    }
+
+    public renderAsMarkdown(displayFileName: string): string {
```
roywill, line 67: It is odd that the file name is a parameter but the content is a member.  Why?

src/embeddedContentFileSystemProvider.ts
```diff
@@ -0,0 +1,310 @@
...
+        const jsonBuffer: Buffer = await new Promise<Buffer>((resolve, reject) => {
+            fs.readFile(parseData.log.fsPath, (err, data) => {
+                err ? reject(err) : resolve(data);
+            });
+        });
+        const jsonMap: JsonMap = require('json-source-map');
+        docMapping = jsonMap.parse(jsonBuffer.toString());
+        return docMapping.data;
+    }
+    /**
```
roywill, line 85: nit newline before?

src/binaryContentRenderer.ts
GabeDeBacker: I should really rename the parameter. This "renderer" knows how to take a "content string" and render it as markdown. The "file" it is representing is really just a "URI" that exists in the SARF log that may not even be reachable. This is really just a "display file name"

---
In reply to: [410408869](https://github.com/microsoft/sarif-vscode-extension/pull/252#discussion_r410408869) [](ancestors = 410408869)

src/binaryContentRenderer.ts
```diff
@@ -0,0 +1,116 @@
...
+        markDownContent =  markDownContent.concat(BinaryContentRenderer.binaryDataMarkdownHeader);
+        for (let bufferIndex: number = 0; bufferIndex < contentBuffer.length; bufferIndex++) {
+            const bufferByte: number = contentBuffer[bufferIndex];
+            // When we hit the start of a new row, add the offset marker.
+            if (bufferIndex % BinaryContentRenderer.bytesPerRow === 0) {
+                markDownContent = markDownContent.concat(`\r\n|0x${bufferIndex < 16 ? '0' : ''}${bufferIndex.toString(16)}`);
+            }
+
+            // Add |0xFF, or |0xFF|\r\n depending on if we are at the end of a row.
+            markDownContent = markDownContent.concat(`|0x${bufferByte < 16 ? '0' : ''}${bufferByte.toString(16)}${((bufferIndex + 1) % BinaryContentRenderer.bytesPerRow === 0) ? '|\r\n' : ''}`);
```
sherryyshi, line 80: >\r\n [](start = 180, length = 4)

is this extension supposed to be cross platform? I would think so...

GabeDeBacker: Yep. I'll see if I can find the correct characters.

---
In reply to: [410424869](https://github.com/microsoft/sarif-vscode-extension/pull/252#discussion_r410424869) [](ancestors = 410424869)

sherryyshi:
consider adding some tests for the new classes?

jeffersonking:
> consider adding some tests for the new classes?

Seconding this. Aside from testing, some expected inputs and outputs would help me understand the desired functionality here better.

jeffersonking:
High level feedback/thoughts:
- The virtual file system is a very welcome approach over temp files.
- I'm hesitant about the binary rendering. I would be more comfortable if this extension was agnostic (or at least minimally involved) in the rendering of files. There are other extensions out there specializing in at business. Likewise, ideally virtual file system should not carry an opinion (or even know) about how these files are rendered. With this PR, that's not the case.

"Minimally involved" in my head would amount to passing a plain text hex dump.

src/embeddedContentFileSystemProvider.ts
```diff
@@ -0,0 +1,332 @@
...
+        const artifact: sarif.Artifact | undefined = run.artifacts[artifactIndex];
+        if (!artifact || !artifact.contents) {
+            return undefined;
+        }
+
+        // We render binary and markdown content as markdown, so
+        // we add the ".md" extension for VSCode to detect that.
+        const uriFileName: string = (artifact.contents.binary  || artifact.contents.rendered?.markdown) ? `${fileName}.md` : fileName;
+        const originalFileNameAsBase64: string = new Buffer(fileName).toString('base64');
+        const logPathAsBase64: string = new Buffer(logPath.toString(/*skipEncoding*/ true), 'UTF8').toString('base64');
```
jeffersonking, line 318: Why Base64 encoding paths versus URI encoding (`encodeURI()`)? The Base64 encoding can still be troublesome during Url.parse().

GabeDeBacker:
Well, I tried about 10 different extensions and, well, none of them worked as I would expect (some not at all). The rendering we are doing here is super minimal and gets the job done.

All that being said, it's up to Michael (or whoever owns this extension :)) to make that call.

I'm on the fence with this. More to the point, I am okay leaving the rendering in, and I'm okay leaving it out.

What I can say from an end-user experience is that seeing the result show up in the viewer, and then clicking on it and having VSCode tell you, sorry dude.. is, well, not great. :)

---
In reply to: [615523970](https://github.com/microsoft/sarif-vscode-extension/pull/252#issuecomment-615523970) [](ancestors = 615523970)

src/embeddedContentFileSystemProvider.ts
```diff
@@ -0,0 +1,328 @@
...
+}
+
+/**
+ * A file system provider that handles embedded content.
+ */
+export class EmbeddedContentFileSystemProvider implements vscode.FileSystemProvider, vscode.Disposable {
+    private disposables: vscode.Disposable[] = [];
+
+    // The URI scheme for the content.
+    private static EmbeddedContentScheme: string = 'sarifEmbeddedContent';
```
GabeDeBacker, line 98: >sarifEmbeddedContent [](start = 52, length = 20)

We should really rename this to sarifArtifactContent because the term "embedded" is not used in the spec.

jeffersonking: Agreed.

src/binaryContentRenderer.ts
```diff
@@ -0,0 +1,211 @@
...
+const localize: nls.LocalizeFunc = nls.loadMessageBundle();
+import * as vscode from 'vscode';
+import * as sarif from 'sarif';
+import MarkdownIt = require("markdown-it");
+
+/**
+ * Class used to render embedded binary content.
+ */
+export class BinaryContentRenderer {
+    private static bytesRenderedPerRow: number = 16;
```
GabeDeBacker, line 15: >16 [](start = 49, length = 2)

Make this user configurable.. No reason not to.

src/binaryContentRenderer.ts
```diff
@@ -0,0 +1,211 @@
...
+import * as nls from 'vscode-nls';
+const localize: nls.LocalizeFunc = nls.loadMessageBundle();
+import * as vscode from 'vscode';
+import * as sarif from 'sarif';
+import MarkdownIt = require("markdown-it");
+
+/**
+ * Class used to render embedded binary content.
+ */
+export class BinaryContentRenderer {
```
GabeDeBacker, line 14: >BinaryContentRenderer [](start = 13, length = 21)

Comment this file way more than it currently is.

src/factories/locationFactory.ts
```diff
@@ -36,6 +39,32 @@ export namespace LocationFactory {
...
+
+                    const artifactContentUri: Uri | undefined = ArtifactContentFileSystemProvider.tryCreateUri(sarifLog, Uri.file(runInfo.sarifFileFullPath), uri, runInfo.runIndex, physLocation.artifactLocation.index);
+                    if (artifactContentUri) {
+                        uri = artifactContentUri;
+                        mappedToLocalPath = true;
+
+                        // For binary content, we can map the byte offset and length to the "markdown" it will be converted to
+                        // for user display.
+                        const binaryContentRenderer: BinaryContentRenderer | undefined = BinaryContentRenderer.tryCreateFromLog(sarifLog, runInfo.runIndex, physLocation.artifactLocation.index);
+                        if (binaryContentRenderer &&
```
jeffersonking, line 57: Internally we need to keep the translated region. However do we show users the original row/col numbers (in the context of the binary), or the translated ones (in the context of markdown)?

GabeDeBacker: We use VSCode's related diagnostic information for that... 

We have many options here.

---
In reply to: [411630299](https://github.com/microsoft/sarif-vscode-extension/pull/252#discussion_r411630299) [](ancestors = 411630299)

src/factories/locationFactory.ts
```diff
@@ -36,6 +39,39 @@ export namespace LocationFactory {
...
+                    if (run && run.artifacts) {
+                        const artifact: sarif.Artifact | undefined = run.artifacts[physLocation.artifactLocation.index];
+                        if (artifact) {
+                            const artifactContent: sarif.ArtifactContent | undefined = artifact.contents;
+                            if (artifactContent) {
+                                fileName = uri.toString(true).substring(uri.toString(true).lastIndexOf('/') + 1);
+
+                                // See if we have a custom renderer for an artifact content object (for example, binary content).
+                                const artifactContentRenderer: ArtifactContentRenderer | undefined = tryCreateRendererForArtifactContent(sarifLog, artifactContent, runInfo.runIndex, physLocation.artifactLocation.index);
+                                const artifactContentUri: Uri | undefined = ArtifactContentFileSystemProvider.tryCreateUri(sarifLog, Uri.file(runInfo.sarifFileFullPath), uri, runInfo.runIndex, physLocation.artifactLocation.index, artifactContentRenderer);
```
GabeDeBacker, line 57: >artifactContentRenderer [](start = 230, length = 23)

I want to change this parameter. The only coupling for URI creation is whether a specific extension is required for the URI (which in the case of binary rendering it needs to be ".md")

jeffersonking:
`artifactContentRendering.ts` -> `artifactContentRenderer.ts`

src/artifactContentRenderers/textArtifactContentRenderer.ts
```diff
@@ -0,0 +1,53 @@
...
+        return new TextArtifactContentRenderer(artifactContents);
+    }
+
+    /**
+     * Renders the contents as mark-down.
+     * @param artifactUri The artifact URI to use as the file-name in the markdown.
+     */
+    public render(artifactUri: vscode.Uri): string {
+        // The text property required to be UTF8 per the spec.
+        return new Buffer(this.contents).toString('utf8');
```
jeffersonking, line 51: You can directly return `this.contents`. Per spec, the incoming data was already utf8.

src/artifactContentRenderers/renderingArtifactContentRenderer.ts
```diff
@@ -0,0 +1,55 @@
...
+
+    /**
+     * Attempts to create an instance of the binary content renderer based on a SARIF log, run index and artifact index
+     * Returns undefined if the artifact contents cannot be found, or the content is not binary content.
+     * @param log The SARIF log.
+     * @param runIndex The run index.
+     * @param artifactIndex The artifact index.
+     */
+    public static tryCreateFromLog(log: sarif.Log, artifactContents: sarif.ArtifactContent, runIndex: number, artifactIndex: number): ArtifactContentRenderer | undefined {
+        if (!artifactContents.rendered || (!artifactContents.rendered.markdown && !artifactContents.rendered.text)) {
```
jeffersonking, line 31: What are your thoughts about being more DRY in general? Such as the conditions here being repeated here and Line 19.

Either check once, or have the other be a try/catch. As it stands, it risks having the two set of conditions fall out of sync.

src/artifactContentRenderers/textArtifactContentRenderer.ts
GabeDeBacker: Need to fix this in the rendering artifact renderer as well :)

---
In reply to: [411721819](https://github.com/microsoft/sarif-vscode-extension/pull/252#discussion_r411721819) [](ancestors = 411721819)

src/artifactContentRenderers/renderingArtifactContentRenderer.ts
jeffersonking: Just realized it's repeated on Line 43 also.

src/artifactContentRenderers/artifactContentRendering.ts
```diff
@@ -0,0 +1,47 @@
...
+     * @param region The region for the artifact content.
+     */
+    rangeFromRegion?(region: sarif.Region): vscode.Range | undefined;
+
+    render(artifactUri: vscode.Uri): string;
+
+    specificUriExtension?: string;
+}
+
+type tryCreateLogFile = (log: sarif.Log, artifactContents: sarif.ArtifactContent, runIndex: number, artifactIndex: number) => ArtifactContentRenderer | undefined;
```
jeffersonking, line 23: `log`, `runIndex`, `artifactIndex` unused.

src/artifactContentRenderers/artifactContentRendering.ts
```diff
@@ -0,0 +1,47 @@
...
+export interface ArtifactContentRenderer {
+    /**
+     * Computes a range from the content it will render.
+     * @param region The region for the artifact content.
+     */
+    rangeFromRegion?(region: sarif.Region): vscode.Range | undefined;
+
+    render(artifactUri: vscode.Uri): string;
+
+    specificUriExtension?: string;
```
jeffersonking, line 20: Also unused.

src/artifactContentRenderers/artifactContentRendering.ts
```diff
@@ -0,0 +1,47 @@
...
+ * Copyright (c) Microsoft Corporation. All Rights Reserved.
+ */
+
+import * as vscode from "vscode";
+import * as sarif from "sarif";
+import { BinaryArtifactContentRenderer } from "./binaryArtifactContentRenderer";
+import { TextArtifactContentRenderer } from "./textArtifactContentRenderer";
+import { RenderingArtifactContentRenderer } from "./renderingArtifactContentRenderer";
+
+export interface ArtifactContentRenderer {
```
jeffersonking, line 11: This design pattern promotes an unnecessary allocation of short-lifespan objects.

src/factories/locationFactory.ts
```diff
@@ -44,15 +80,16 @@ export namespace LocationFactory {
...
-
-                fileName = uri.toString(true).substring(uri.toString(true).lastIndexOf('/') + 1);
             }
         }
 
-        if (physLocation && physLocation.region) {
-            parsedRange = parseRange(physLocation.region);
-            message = Utilities.parseSarifMessage(physLocation.region.message);
+        if (!parsedRange) {
+            if (physLocation && physLocation.region) {
```
jeffersonking, line 89: Combine predicates.

src/factories/locationFactory.ts
```diff
@@ -36,6 +39,39 @@ export namespace LocationFactory {
             uri = Utilities.combineUriWithUriBase(physLocation.artifactLocation.uri, uriBase);
 
             if (uri) {
+                // Check for artifact content and create our artifact content URI.
+                // In order to have artifact content, the log must have a valid artifact index
+                // (represented by artifact location index) and the "contents" property of the artifact
+                // must be defined and have one of the "binary, text, or rendered" properties set.
+                if (physLocation.artifactLocation.index !== undefined) {
+                    const run: sarif.Run | undefined = sarifLog.runs[runInfo.runIndex];
+                    if (run && run.artifacts) {
```
jeffersonking, line 48: Optional chaining.

src/factories/locationFactory.ts
```diff
@@ -36,6 +39,39 @@ export namespace LocationFactory {
...
+                        const artifact: sarif.Artifact | undefined = run.artifacts[physLocation.artifactLocation.index];
+                        if (artifact) {
+                            const artifactContent: sarif.ArtifactContent | undefined = artifact.contents;
+                            if (artifactContent) {
+                                fileName = uri.toString(true).substring(uri.toString(true).lastIndexOf('/') + 1);
+
+                                // See if we have a custom renderer for an artifact content object (for example, binary content).
+                                const artifactContentRenderer: ArtifactContentRenderer | undefined = tryCreateRendererForArtifactContent(sarifLog, artifactContent, runInfo.runIndex, physLocation.artifactLocation.index);
```
jeffersonking, line 56: Why is this instantiated at this (outer) scope?

jeffersonking:
For me, the binary files are opening as textual markdown rather than rendered markdown. Is there a editor setting I need to change? I don't see any.

src/artifactContentFileSystemProvider.ts
```diff
@@ -0,0 +1,333 @@
...
+
+        return {
+            dispose: () => { fs.unwatchFile(parsedUriData.log.fsPath, watcher); }
+        };
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public async stat(uri: vscode.Uri): Promise<vscode.FileStat> {
```
jeffersonking, line 160: Returning dummy values seems to make no difference for our scenarios. Why are implementing this fully?

src/artifactContentFileSystemProvider.ts
```diff
@@ -0,0 +1,333 @@
...
+        const run: sarif.Run | undefined = sarifLog.runs[runIndex];
+        if (!run || !run.artifacts) {
+            return undefined;
+        }
+        const artifact: sarif.Artifact | undefined = run.artifacts[artifactIndex];
+        if (!artifact || !artifact.contents) {
+            return undefined;
+        }
+
+        let uriPath: string;
```
jeffersonking, line 271: Why is `uriPath` needed? It is ignored downstream, including by `parseUri()`.

src/artifactContentFileSystemProvider.ts
```diff
@@ -0,0 +1,333 @@
...
+        }
+
+        const encodedUriData: EncodedUriData = {
+            artifactIndex,
+            runIndex,
+            log: logPath.toString(/*skipEncoding*/ true),
+            artifactUri: artifactUri.toString(/*skpEncoding*/ true)
+        };
+
+        const encodedUriDataAsBase64: string = new Buffer(JSON.stringify(encodedUriData)).toString('base64');
```
jeffersonking, line 286: Could we have the URI as human readable as possible for debugging purposes? Index values (numeric) are URI safe and can be transparent. String values can be URI encoded for maximum readability. You can use the path split and store these values.

src/artifactContentFileSystemProvider.ts
GabeDeBacker: uriPath (which is the end of the URI before the path)  contains the file extension that VSCode will use to detect the file-type. I've added a comment.

---
In reply to: [411818093](https://github.com/microsoft/sarif-vscode-extension/pull/252#discussion_r411818093) [](ancestors = 411818093)

src/factories/locationFactory.ts
GabeDeBacker: I would love to except the type-script compiler doesn't appear to like if checks like this:

if (sarifLog.runs[runInfo.runIndex]?.artifacts && 
sarifLog.runs[runInfo.runIndex].artifacts[physLocation.artifactLocation.index])

 "resource": "/d:/Gabriel/Documents/Repos/sarif-vscode-extension/src/factories/locationFactory.ts",
 "owner": "typescript",
 "message": "Object is possibly 'undefined'.",
 "owner": "typescript",

Decided not to fight it.

---
In reply to: [411738990](https://github.com/microsoft/sarif-vscode-extension/pull/252#discussion_r411738990) [](ancestors = 411738990)

GabeDeBacker:
That is correct. VSCode does not apply the diagnostic ranges to the preview, so opening to the textual representation is actually better than going to the mark-down.

In the future, if we decide to get even more sophisticated with the markdowns, we could but highlights in the markdown with tooltips for the issues. That can be a future enhancement.

I added an option to open markdown content preview.


---
In reply to: [616873222](https://github.com/microsoft/sarif-vscode-extension/pull/252#issuecomment-616873222) [](ancestors = 616873222)

src/artifactContentRenderers/artifactContentRendering.ts
GabeDeBacker: It is used to create the URI for VSCode to detect the language. It is referenced in the location factory as well as the artifact content file system provider.

I forgot to comment these functions, so I've done so.

---
In reply to: [411732865](https://github.com/microsoft/sarif-vscode-extension/pull/252#discussion_r411732865) [](ancestors = 411732865)

src/artifactContentRenderers/artifactContentRendering.ts
GabeDeBacker: This is defining a type for line 25, it isn't implementing a function

---
In reply to: [411728013](https://github.com/microsoft/sarif-vscode-extension/pull/252#discussion_r411728013) [](ancestors = 411728013)

src/artifactContentRenderers/artifactContentRendering.ts
GabeDeBacker: It's not creating any objects (unless I'm missing something). There is an array of functions (that are static in their implementation). No objects are created unless the renderer supports the artifact content.

In effect, it is simply splitting the "if check" that used to exist in one function into three smaller functions.

---
In reply to: [411735831](https://github.com/microsoft/sarif-vscode-extension/pull/252#discussion_r411735831) [](ancestors = 411735831)

src/artifactContentFileSystemProvider.ts
GabeDeBacker: This pattern (using base64) was used explicitly so I did not have to implement my own parsing.

Writing your own string parsing should at all costs be avoided as it is super easy to miss edge cases and in some cases cause security issues.

Using a well tested encoder\decoder for this is a better choice.

---
In reply to: [411819043](https://github.com/microsoft/sarif-vscode-extension/pull/252#discussion_r411819043) [](ancestors = 411819043)

src/artifactContentFileSystemProvider.ts
GabeDeBacker: We can remove the size calculation for sure.

VSCode will use the times to determine if it can show a "cached" view (on re-start) of content it may have.

---
In reply to: [411806880](https://github.com/microsoft/sarif-vscode-extension/pull/252#discussion_r411806880) [](ancestors = 411806880)

src/artifactContentRenderers/renderingArtifactContentRenderer.ts
```diff
@@ -0,0 +1,56 @@
...
+ */
+
+import * as vscode from 'vscode';
+import * as sarif from 'sarif';
+import { ArtifactContentRenderer } from './artifactContentRenderer';
+
+/**
+ * Class used to render artifact content objects with text.
+ */
+export class RenderingArtifactContentRenderer implements ArtifactContentRenderer {
```
GabeDeBacker, line 12: >RenderingArtifactContentRenderer [](start = 13, length = 32)

Name this to rendered

GabeDeBacker: Filename as well

---
In reply to: [412388852](https://github.com/microsoft/sarif-vscode-extension/pull/252#discussion_r412388852) [](ancestors = 412388852)

src/artifactContentRenderers/artifactContentRenderer.ts
```diff
@@ -0,0 +1,59 @@
...
+
+    /**
+     * Specifies a specific file extension needed for VSCode to detect the
+     * document type. For example, makrdown => '.md'. The extension
+     * should include any necessary periods.
+     */
+    specificUriExtension?: string;
+}
+
+type TryCreateLogFile = (log: sarif.Log, artifactContent: sarif.ArtifactContent, runIndex: number, artifactIndex: number) => ArtifactContentRenderer | undefined;
```
GabeDeBacker, line 35: >, runIndex: number, artifactIndex: number [](start = 79, length = 41)

Since we are passing the artifact content, these indices are not needed anymore.

src/artifactContentFileSystemProvider.ts
jeffersonking: Propose: sarif:/logUri/runIndex/artifactIndex/file.md

jeffersonking: Noting parsing as a potential security risk (Gabe).

jeffersonking:
> That is correct. VSCode does not apply the diagnostic ranges to the preview, so opening to the textual representation is actually better than going to the mark-down.
> 
> In the future, if we decide to get even more sophisticated with the markdowns, we could but highlights in the markdown with tooltips for the issues. That can be a future enhancement.
> 
> I added an option to open markdown content preview.
> 
> In reply to: [616873222](https://github.com/microsoft/sarif-vscode-extension/pull/252#issuecomment-616873222) [](ancestors = 616873222)

If this leads us to just return plain text, we should reconsider `TextDocumentContentProvider` which should be simpler.

jeffersonking:
Reviewed "in person". Would like to see some of the changes we discussed implemented, but also don't want to block your progress.

src/artifactContentFileSystemProvider.ts
GabeDeBacker: We can consider this based on feedback from users since these URIs show up when you hover over the tab. Having the URI most relevant to the artifact is, in my opinion, better

---
In reply to: [412415505](https://github.com/microsoft/sarif-vscode-extension/pull/252#discussion_r412415505) [](ancestors = 412415505)

src/factories/locationFactory.ts
GabeDeBacker: We can open an issue later if we deem it to be necessary.

---
In reply to: [411702385](https://github.com/microsoft/sarif-vscode-extension/pull/252#discussion_r411702385) [](ancestors = 411702385,411630299)

Issue #247, opened 2020-04-15
Sarif Viewer opens on any JSON file open


GabeDeBacker:
The viewer opens every time you open a JSON file, regardless if it is SARIF or not.

Issue #143, opened 2019-02-26
Sarif Explorer is opening when extension settings change


rscrivens:
We should not be launching the Sarif Explorer when settings change. Only apply the settings if the Sarif Explorer is already open.

PR #253, opened 2020-04-21
Stop opening explorer at weird times

GabeDeBacker:
This change addresses issues #247 and issue #143. The root cause of these issues is that the extension would activate and show the web-view panel any time a message was requested to be posted to it.

This simply isn't required.

So the change adds options for how the post message should be processed:
```typescript
export type SendMessageOptions = 'Always Open' | 'Only if open' | 'Should already be open';
```
Since I was already changing the code to optionally open the viewer based on the above type, I finished up plumbing through "Show viewer when no results" from the API to the explorer controller.


package-lock.json
```diff
@@ -31,31 +31,31 @@
             }
         },
         "@microsoft/sarif-multitool": {
-            "version": "2.2.2",
-            "resolved": "https://registry.npmjs.org/@microsoft/sarif-multitool/-/sarif-multitool-2.2.2.tgz",
-            "integrity": "sha512-mid4xKQv5QkrbZtH+V0Tlwl3aCneZtAEoggXqY7klwuza1Zn1uI7CkRqq8aDiv+fWR/a6acAbagEogwzfRVY8A==",
+            "version": "2.2.3",
```
GabeDeBacker, line 34: This once again is because the package-lock json file in master has not been updated yet.

src/explorerController.ts
```diff
@@ -132,11 +137,7 @@ export class ExplorerController implements Disposable {
 
             case MessageType.ExplorerLoaded:
                 if (this.resultsListData) {
-                    const webViewMessage: WebviewMessage = {
-                        data: JSON.stringify(this.resultsListData),
-                        type: MessageType.ResultsListDataSet
-                    };
-                    this.sendMessage(webViewMessage);
+                    this.setResultsListData(this.resultsListData, 'Only if open');
```
GabeDeBacker, line 140: >'Only if open' [](start = 66, length = 14)

This should be "should already be open"

src/extension.ts
```diff
@@ -18,191 +18,238 @@ import { Utilities } from "./utilities";
...
-    void readOpenedDocuments(logReader, diagnosticCollection);
-
-    return new ApiImpl(logReader, diagnosticCollection);
+    return new SarifExtension(context);
 }
 
 /**
  * Clean up extension if it gets deactivated
  */
 export function deactivate(): void {
```
GabeDeBacker, line 34: >deactivate [](start = 16, length = 10)

Verify that this disposes the extension properly.

package.json
```diff
@@ -164,6 +164,11 @@
                             "type": "boolean"
                         }
                     }
+                },
+                "sarif-viewer.explorer.openWhenNoResults": {
+                    "description": "%sarif-viewer.explorer.openWhenNoResults.description%",
+                    "type": "boolean",
+                    "default": false
```
GabeDeBacker, line 171: >false [](start = 31, length = 5)

This should be true.

src/explorerController.ts
```diff
@@ -253,18 +264,37 @@ export class ExplorerController implements Disposable {
...
+        this.sendMessage({data: dataString, type: MessageType.NewDiagnostic}, 'Always Open');
     }
 
     /**
      * Handles sending a message to the webview
      * @param message Message to send, message has a type and data
+     * @param options Options specifying whether to show the explorer view.
      */
-    private sendMessage(message: WebviewMessage): void {
+    private sendMessage(message: WebviewMessage, options: SendMessageOptions): void {
```
GabeDeBacker, line 275: >sendMessage [](start = 12, length = 11)

rename this to post message

src/explorerController.ts
```diff
@@ -14,6 +14,8 @@ import { SarifViewerVsCodeDiagnostic } from "./sarifViewerDiagnostic";
 import { Utilities } from "./utilities";
 import { SVDiagnosticCollection } from "./svDiagnosticCollection";
 
+export type SendMessageOptions = 'Always Open' | 'Only if open' | 'Should already be open';
```
GabeDeBacker, line 17: >SendMessageOptions [](start = 12, length = 18)

Change to PostMessageOptions

GabeDeBacker, line 17: >SendMessageOptions [](start = 12, length = 18)

Comment the meaning of these thoroughly

src/extension.ts
GabeDeBacker: Confirmed. Everything disposes correctly.

---
In reply to: [412425847](https://github.com/microsoft/sarif-vscode-extension/pull/253#discussion_r412425847) [](ancestors = 412425847)

Issue #232, opened 2020-04-10
Keyboard navigation in results list, well, doesn't work


GabeDeBacker:
While prototyping a fix for #231, which will leave "focus" in the explorer view, you quickly discover that while focus stays in the results list viewer, you cannot use the arrow keys to navigate between them.

Issue #231, opened 2020-04-10
Add an option for taking focus to source when moving between results


GabeDeBacker:
When moving between results in the "explorer" viewer, some users seem to prefer that it take them to the source code AND put focus on the editor, while others prefer opening the source, but leaving focus in the editor.

Seems legitimate and an easy option to add.

PR #254, opened 2020-04-22
Improve keyboard navgiation in result list

GabeDeBacker:
The keyboard navigation was pretty much non-existent in the results list of the explorer. We need at least "something" to get this working. This address issue #232 

This implements the bare minimum of keyboard navigation through the results list.

As you arrow "down" or "right" throw the results list, it will expand the groups and select the next result.

As you arrow "up" it will expand the group and select the previous result.

As you arrow "left" it will collapse the group that the selection is moving from and then select the previous result (in the previous group) unless it is the top group.

This also improves the "expanding" or "collapsing" of a group by using the jquery selector to find the groups that need to change rather than enumerating through all the results (which improves perf a bit).

@jeffersonking Is currently investigating a new way to implement the explorer view (which is more modern) so this fix is to address current concerns.

### DIFF
--- a/package.nls.json
+++ b/package.nls.json
@@ -125,6 +125,5 @@
     "sarif-viewer.explorer.openWhenNoResults.description": "Indicates whether to open the explorer when there are no results in the log.",
     "sarif-viewer.binaryArtifactContent.bytesPerRow.description": "When rendering binary artifact content, determines the number of bytes displayed in a row.",
     "sarif-viewer.binaryArtifactContent.viewAsMarkdown.description": "When enabled, renders binary artifact content as markdown.",
-    "sarif-viewer.openMarkdownPreview.description": "When viewing artifact content containing markdown, show the markdown preview.",
-    "sarif-viewer.explorer.preserveKeyboardFocus.description": "When selecting results in the explorer, preserve keyboard focus in the results explorer. Otherwise, move focus to the result artifact."
+    "sarif-viewer.openMarkdownPreview.description": "When viewing artifact content containing markdown, show the markdown preview."
 }
--- a/package.nls.json
+++ b/package.nls.json
@@ -125,5 +125,6 @@
     "sarif-viewer.explorer.openWhenNoResults.description": "Indicates whether to open the explorer when there are no results in the log.",
     "sarif-viewer.binaryArtifactContent.bytesPerRow.description": "When rendering binary artifact content, determines the number of bytes displayed in a row.",
     "sarif-viewer.binaryArtifactContent.viewAsMarkdown.description": "When enabled, renders binary artifact content as markdown.",
-    "sarif-viewer.openMarkdownPreview.description": "When viewing artifact content containing markdown, show the markdown preview."
+    "sarif-viewer.openMarkdownPreview.description": "When viewing artifact content containing markdown, show the markdown preview.",
+    "sarif-viewer.explorer.preserveKeyboardFocus.description": "When selecting results in the explorer, preserver keyboard focus in the results explorer. Otherwise, move focus to the result artifact."
 }
--- a/package.nls.json
+++ b/package.nls.json
@@ -126,5 +126,5 @@
     "sarif-viewer.binaryArtifactContent.bytesPerRow.description": "When rendering binary artifact content, determines the number of bytes displayed in a row.",
     "sarif-viewer.binaryArtifactContent.viewAsMarkdown.description": "When enabled, renders binary artifact content as markdown.",
     "sarif-viewer.openMarkdownPreview.description": "When viewing artifact content containing markdown, show the markdown preview.",
-    "sarif-viewer.explorer.preserveKeyboardFocus.description": "When selecting results in the explorer, preserver keyboard focus in the results explorer. Otherwise, move focus to the result artifact."
+    "sarif-viewer.explorer.preserveKeyboardFocus.description": "When selecting results in the explorer, preserve keyboard focus in the results explorer. Otherwise, move focus to the result artifact."
 }
--- a/src/explorer/resultsList.ts
+++ b/src/explorer/resultsList.ts
@@ -612,6 +612,19 @@ export class ResultsList {
             throw new Error("Expected to have collapse group.");
         }
 
+        // If we are passed a row instead of a group, then go find the correct group.
+        // This is useful for handling keyboard navigation.
+        if (row.classList.contains('listtablerow')) {
+            const groups: NodeListOf<Element> = document.querySelectorAll(`#resultslisttable > tbody > .listtablegroup[data-group='${row.dataset.group}']`);
+            if (groups.length !== 1) {
+                throw new Error('Could not find expected group');
+            }
+
+            this.setToggleStateForGroupOfResults(<HTMLTableRowElement>groups[0], newState);
+
+            return;
+        }
+
         if (newState === ToggleState.expanded && row.classList.contains(ToggleState.collapsed)) {
             row.classList.replace(ToggleState.collapsed, ToggleState.expanded);
             this.collapsedGroups.splice(this.collapsedGroups.indexOf(row.dataset.group), 1);
@@ -738,18 +751,11 @@ export class ResultsList {
      * @param keydownEvent Keyboard event
      */
     private onResultsListKeyDown(keydownEvent: JQuery.KeyDownEvent): void  {        
-        enum KeyCodes {
-            LeftArrow = 37,
-            UpArrow = 38,
-            RightArrow = 39,
-            DownArrow = 40
-        }
-
         let direction: 'Previous' | 'Next' | undefined;
-        if (keydownEvent.keyCode === KeyCodes.LeftArrow || keydownEvent.keyCode === KeyCodes.UpArrow) {
+        if (keydownEvent.key === 'ArrowLeft' || keydownEvent.key === 'ArrowUp') {
             direction = 'Previous';
             
-        } else if (keydownEvent.keyCode === KeyCodes.RightArrow || keydownEvent.keyCode == KeyCodes.DownArrow) {
+        } else if (keydownEvent.key === 'ArrowRight' || keydownEvent.key == 'ArrowDown') {
             direction = 'Next';
         } else {
             return;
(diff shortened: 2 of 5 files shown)


src/explorer/resultsList.ts
```diff
@@ -177,7 +179,7 @@ export class ResultsList {
             // Collapse All button
             const collapseAllButton: HTMLSpanElement = <HTMLSpanElement>expandAllButton.cloneNode();
             collapseAllButton.textContent = "-";
-            collapseAllButton.title = "Collapse all groups";
+            collapseAllButton.title = "Collapse all s";
```
sherryyshi, line 182: > s [](start = 51, length = 2)

s?

src/explorer/resultsList.ts
```diff
@@ -748,4 +732,64 @@ export class ResultsList {
...
+     */
+    private onResultsListKeyDown(keydownEvent: JQuery.KeyDownEvent): void  {        
+        enum KeyCodes {
+            LeftArrow = 37,
+            UpArrow = 38,
+            RightArrow = 39,
+            DownArrow = 40
+        }
+
+        let direction: 'Previous' | 'Next' | undefined;
```
sherryyshi, line 748: >undefined [](start = 45, length = 9)

not needed?

src/explorer/resultsList.ts
```diff
@@ -748,4 +732,64 @@ export class ResultsList {
...
+        if (keydownEvent.keyCode === KeyCodes.LeftArrow || keydownEvent.keyCode === KeyCodes.UpArrow) {
+            direction = 'Previous';
+            
+        } else if (keydownEvent.keyCode === KeyCodes.RightArrow || keydownEvent.keyCode == KeyCodes.DownArrow) {
+            direction = 'Next';
+        }
+
+        if (!direction) {
```
sherryyshi, line 756: else

src/explorer/resultsList.ts
```diff
@@ -748,4 +732,64 @@ export class ResultsList {
...
+        }
+
+        const curSelected: HTMLCollectionOf<Element> | undefined = document.getElementsByClassName("listtablerow selected");
+        if (!curSelected || curSelected.length !== 1) {
+            return;
+        }
+
+        let currentSelectedElement: Element | undefined = curSelected[0];
+        while (currentSelectedElement) {
+            const potentialElementToSelect: Element | null = direction === 'Next' ? currentSelectedElement.nextElementSibling : currentSelectedElement.previousElementSibling;
```
sherryyshi, line 767: >null [](start = 54, length = 4)

just making sure - is this actually `null` or should it be `undefined`?

src/explorer/resultsList.ts
GabeDeBacker: hmm, oops

---
In reply to: [413106683](https://github.com/microsoft/sarif-vscode-extension/pull/254#discussion_r413106683) [](ancestors = 413106683)

jeffersonking:
Could we discuss the config `preserveKeyboardFocus`? How many people do we know want it `false`?

Considering existing UI like the Problems panel and Explorer, the predominant pattern is to preserve keyboard focus. I don't think those are configurable either.

src/explorer/resultsList.ts
GabeDeBacker: Yes, it really is null.

---
In reply to: [413112674](https://github.com/microsoft/sarif-vscode-extension/pull/254#discussion_r413112674) [](ancestors = 413112674)

GabeDeBacker:
Don't disagree with that statement at all. The ONLY reason I made it a setting is that the existing behavior was to NOT preserve focus, so I made an option for anyone that may want that behavior.

Let me check on any VSCode settings.

---
In reply to: [617885676](https://github.com/microsoft/sarif-vscode-extension/pull/254#issuecomment-617885676) [](ancestors = 617885676)

package.nls.json
```diff
@@ -125,5 +125,6 @@
     "sarif-viewer.explorer.openWhenNoResults.description": "Indicates whether to open the explorer when there are no results in the log.",
     "sarif-viewer.binaryArtifactContent.bytesPerRow.description": "When rendering binary artifact content, determines the number of bytes displayed in a row.",
     "sarif-viewer.binaryArtifactContent.viewAsMarkdown.description": "When enabled, renders binary artifact content as markdown.",
-    "sarif-viewer.openMarkdownPreview.description": "When viewing artifact content containing markdown, show the markdown preview."
+    "sarif-viewer.openMarkdownPreview.description": "When viewing artifact content containing markdown, show the markdown preview.",
+    "sarif-viewer.explorer.preserveKeyboardFocus.description": "When selecting results in the explorer, preserver keyboard focus in the results explorer. Otherwise, move focus to the result artifact."
```
roywill, line 129: Either stick "Life" in front of this or remove the "r" if you mean preserve

src/extension.ts
```diff
@@ -90,6 +91,9 @@ class SarifExtension implements Api {
         this.logReader = new LogReader();
         context.subscriptions.push(this.logReader);
 
+        // Register our file system provider.
+        context.subscriptions.push(new ArtifactContentFileSystemProvider());
```
roywill, line 95: What does this have to do with keyboard navigation?

src/explorer/resultsList.ts
```diff
@@ -748,4 +732,62 @@ export class ResultsList {
...
+    private onResultsListKeyDown(keydownEvent: JQuery.KeyDownEvent): void  {        
+        enum KeyCodes {
+            LeftArrow = 37,
+            UpArrow = 38,
+            RightArrow = 39,
+            DownArrow = 40
+        }
+
+        let direction: 'Previous' | 'Next' | undefined;
+        if (keydownEvent.keyCode === KeyCodes.LeftArrow || keydownEvent.keyCode === KeyCodes.UpArrow) {
```
jeffersonking, line 749: `.keyCode` is deprecated. `.key` is preferred.

src/extension.ts
GabeDeBacker: Ah.. nothing. This was removed accidentally when I merged the last PR to master.

Result => artifact content objects (like binary content) fail to render (oops).

---
In reply to: [413139862](https://github.com/microsoft/sarif-vscode-extension/pull/254#discussion_r413139862) [](ancestors = 413139862)

roywill:
What does the "Spec" say the behavior was supposed to be?  That would need to be fixed first.  As a result, what Gabriel has is a way to maintain previous behavior.

---
In reply to: [617890878](https://github.com/microsoft/sarif-vscode-extension/pull/254#issuecomment-617890878) [](ancestors = 617890878,617885676)

jeffersonking:
> What does the "Spec" say the behavior was supposed to be? That would need to be fixed first. As a result, what Gabriel has is a way to maintain previous behavior.
> 
> In reply to: [617890878](https://github.com/microsoft/sarif-vscode-extension/pull/254#issuecomment-617890878) [](ancestors = 617890878,617885676)



> Don't disagree with that statement at all. The ONLY reason I made it a setting is that the existing behavior was to NOT preserve focus, so I made an option for anyone that may want that behavior.
> 
> Let me check on any VSCode settings.
> 
> In reply to: [617885676](https://github.com/microsoft/sarif-vscode-extension/pull/254#issuecomment-617885676) [](ancestors = 617885676)

Could we

src/explorer/resultsList.ts
GabeDeBacker: Yep. Just noticed that... Looking into this

---
In reply to: [413140401](https://github.com/microsoft/sarif-vscode-extension/pull/254#discussion_r413140401) [](ancestors = 413140401)

jeffersonking:
> Don't disagree with that statement at all. The ONLY reason I made it a setting is that the existing behavior was to NOT preserve focus, so I made an option for anyone that may want that behavior.
> 
> Let me check on any VSCode settings.
> 
> In reply to: [617885676](https://github.com/microsoft/sarif-vscode-extension/pull/254#issuecomment-617885676) [](ancestors = 617885676)

Could we double-check this claim: "that the existing behavior was to NOT preserve focus"?

I have 2.14.0 (from the Marketplace) installed and I am not seeing that.

Furthermore, #231 is titled "Add an option for taking focus to source when moving between results". Implying focus was *not* being taken to source and thus being preserved in the past.

src/explorer/resultsList.ts
```diff
@@ -748,4 +732,62 @@ export class ResultsList {
...
+            direction = 'Previous';
```
jeffersonking, line 750: How was it decided that Left Arrow and Right Arrow are also mapped to previous and next? This is inconsistent with common UI patterns.

src/explorer/resultsList.ts
```diff
@@ -748,4 +732,62 @@ export class ResultsList {
...
+        let currentSelectedElement: Element | undefined = curSelected[0];
+        while (currentSelectedElement) {
+            const potentialElementToSelect: Element | null = direction === 'Next' ? currentSelectedElement.nextElementSibling : currentSelectedElement.previousElementSibling;
+
+            // If the element is not a table row, then we are done.
+            if (!potentialElementToSelect || potentialElementToSelect.tagName.localeCompare('tr', 'root', {sensitivity: 'base'}) !== 0) {
+                break;
+            }
+
+            const rowElement: HTMLTableRowElement = (<HTMLTableRowElement>potentialElementToSelect);
```
jeffersonking, line 778: Extra parens.

src/explorer/resultsList.ts
```diff
@@ -748,4 +732,62 @@ export class ResultsList {
...
+        if (!curSelected || curSelected.length !== 1) {
+            return;
+        }
+
+        let currentSelectedElement: Element | undefined = curSelected[0];
+        while (currentSelectedElement) {
+            const potentialElementToSelect: Element | null = direction === 'Next' ? currentSelectedElement.nextElementSibling : currentSelectedElement.previousElementSibling;
+
+            // If the element is not a table row, then we are done.
+            if (!potentialElementToSelect || potentialElementToSelect.tagName.localeCompare('tr', 'root', {sensitivity: 'base'}) !== 0) {
```
jeffersonking, line 774: `potentialElementToSelect?.tagName !== 'TR'`

src/explorer/resultsList.ts
GabeDeBacker: That may be true, but it's good enough for now until your work is done.

---
In reply to: [413164231](https://github.com/microsoft/sarif-vscode-extension/pull/254#discussion_r413164231) [](ancestors = 413164231)

jeffersonking:
Trying out this change, but blocked by #255 where I can't click on the result to begin with.

GabeDeBacker:
Interesting. I'll look into #255. I'm using SARIF files where it either contains artifact content(embedded) or I have the source files.

---
In reply to: [617936956](https://github.com/microsoft/sarif-vscode-extension/pull/254#issuecomment-617936956) [](ancestors = 617936956)

jeffersonking:
> Interesting. I'll look into #255. I'm using SARIF files where it either contains artifact content(embedded) or I have the source files.
> 
> In reply to: [617936956](https://github.com/microsoft/sarif-vscode-extension/pull/254#issuecomment-617936956) [](ancestors = 617936956)

Ok, I'll temp test with those types of logs.

GabeDeBacker:
Went and looked at the "original" code, I changed the behavior, so removing the setting and retaining original behavior.

---
In reply to: [617894660](https://github.com/microsoft/sarif-vscode-extension/pull/254#issuecomment-617894660) [](ancestors = 617894660,617890878,617885676)

GabeDeBacker:
Went and looked at the "original" code, I changed the behavior, so removing the setting and retaining original behavior.

---
In reply to: [617904822](https://github.com/microsoft/sarif-vscode-extension/pull/254#issuecomment-617904822) [](ancestors = 617904822)

Issue #256, opened 2020-04-22
Results Info panel does not always update.


GabeDeBacker:
Given a simple C++ file and SARIF file, the results show fine, but selecting different results often results in the "result info" panel not updating.

```cpp
#include <stdio.h>
#include <string>

int main(int argc, __in_ecount(argc) char* argv[])
{
    std::string hello("Hello, Island!\n");

    printf(hello.c_str());
    return 0;
}
```
SARIF
```json
{
  "version": "2.0.0",
  "$schema": "http://json.schemastore.org/sarif-2.0.0",
  "runs": [
    {
      "results": [
        {
          "ruleId": "26487",
          "message": {
            "text": "Don't return a pointer '(*argv)' that may be invalid (lifetime.4)."
          },
          "locations": [
            {
              "physicalLocation": {
                "fileLocation": {
                  "uri": "file:///e:/tb0/src/sdktools/island/island.cpp"
                },
                "region": {
                  "startLine": 4,
                  "startColumn": 5,
                  "endLine": 4,
                  "endColumn": 5
                }
              },
              "fullyQualifiedLogicalName": "main"
            }
          ],
          "codeFlows": [
            {
              "threadFlows": [
                {
                  "locations": [
                    {
                      "step": 0,
                      "location": {
                        "physicalLocation": {
                          "fileLocation": {
                            "uri": "file:///e:/tb0/src/sdktools/island/island.cpp"
                          },
                          "region": {
                            "startLine": 4,
                            "startColumn": 5
                          }
                        }
                      },
                      "importance": "important"
                    }
                  ]
                }
              ]
            }
          ]
        },
        {
          "ruleId": "26485",
          "message": {
            "text": "Expression 'argv': No array to pointer decay (bounds.3)."
          },
          "locations": [
            {
              "physicalLocation": {
                "fileLocation": {
                  "uri": "file:///e:/tb0/src/sdktools/island/island.cpp"
                },
                "region": {
                  "startLine": 4,
                  "startColumn": 5,
                  "endLine": 4,
                  "endColumn": 5
                }
              }
            }
          ]
        }
      ],
      "tool": {
        "name": "PREfast",
        "fullName": "PREfast Code Analysis",
        "version": "14.23.28200.2",
        "fileVersion": "14.00.28200.02"
      },
     "files": {
        "file:///e:/tb0/src/sdktools/island/island.cpp": {
          "roles": [
            "analysisTarget",
            "resultFile"
          ]
        }
      },
[...]

Issue #255, opened 2020-04-22
Explorer: Unable to click on a file (which is unmapped).


jeffersonking:
Clicking on the unmapped result raises the mapping dialog. If I cancel, it's not remembering. Next time I click, I get the dialog again. So effectively I can't click the result.

PR #257, opened 2020-04-22
File mapping strikes again.. (no active result set when file mapping hasn't been completed)

GabeDeBacker:
This adress issues #256 and #255 
There were a couple of hidden things going on.

When a result was clicked in the explorer view, the "result list controller" receives a message. The result list controller needed to set what the rest of the extension believes is the "active diagnostic" (selected result) through the rest of the system. It was not doing that at all (something I like regressed).

After doing that, I ran into the fact that the source links did not work clicked on. Fixed that too.

There were 2 issues there, the first was that there was a Tslint (ignore) in the code, which, well, shouldn't have been ignored as it was causing an exception since "external" really was undefined. The second was that when the source link was clicked on, we assumed that it had already been "mapped" by the file-mapper, which was the wrong assumption.

### DIFF
--- a/src/common/interfaces.d.ts
+++ b/src/common/interfaces.d.ts
@@ -88,6 +88,12 @@ export interface Location {
      * Event that is fired when the location is mapped..
      */
     locationMapped: Event<Location>;
+
+    /**
+     * When a location is converted to "JSON", we covert the URI to a string and store it here
+     * so the web-view can use it since it does not have easy access to vscode's node APIs.
+     */
+    readonly uriAsString?: string;
 }
 
 
--- a/src/common/interfaces.d.ts
+++ b/src/common/interfaces.d.ts
@@ -90,7 +90,7 @@ export interface Location {
     locationMapped: Event<Location>;
 
     /**
-     * When a location is converted to "JSON", we covert the URI to a string and store it here
+     * When a location is converted to "JSON", we convert the URI to a string and store it here
      * so the web-view can use it since it does not have easy access to vscode's node APIs.
      */
     readonly uriAsString?: string;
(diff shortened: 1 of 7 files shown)


src/common/interfaces.d.ts
```diff
@@ -88,6 +88,12 @@ export interface Location {
      * Event that is fired when the location is mapped..
      */
     locationMapped: Event<Location>;
+
+    /**
+     * When a location is converted to "JSON", we covert the URI to a string and store it here
```
roywill, line 93: >covert [](start = 50, length = 6)

convert?

src/explorer/webview.ts
```diff
@@ -346,9 +346,8 @@ export class ExplorerWebview {
             if (options.tooltip === undefined) {
                 if (options.message !== undefined) {
                     options.tooltip = options.message;
-                } else if (options.location.uri !== undefined) {
-                    // @ts-ignore external exist on the webview side
-                    options.tooltip = options.location.uri.external.replace("%3A", ":");
+                } else if (options.location.uriAsString !== undefined) {
```
roywill, line 349: >.location.uriAsString !== undefined [](start = 34, length = 35)

Why does below just use options.location.uriAsString?

GabeDeBacker: They should indeed both be checking for === undefined.

---
In reply to: [413326716](https://github.com/microsoft/sarif-vscode-extension/pull/257#discussion_r413326716) [](ancestors = 413326716)

src/explorer/webview.ts
```diff
@@ -843,18 +841,16 @@ export class ExplorerWebview {
      */
     private createSourceLink(location: Location, linkText: string): HTMLAnchorElement | undefined {
         let sourceLink: HTMLAnchorElement | undefined;
-        if (location.uri && location.range) {
-            // @ts-ignore external exist on the webview side
-            const file: string = location.uri.external.replace("%3A", ":");
+        if (location.uriAsString && location.range) {
```
GabeDeBacker, line 844: >uriAsString [](start = 21, length = 11)

Need to check against undefined, not just an empty string.

PR #267, opened 2020-04-25
Implement update mechanism for Inisiders Channel

GabeDeBacker:
Now that we have Insiders releases being created on GitHub's releases automatically through our ADO build pipeline, we can now create an update mechanisms for those insiders.

The idea is pretty simple and was adapted from the [Microsoft C/C++ extension](https://github.com/microsoft/vscode-cpptools).

Use the GitHub APIs to read the releases and look at the tags (which are the version numbers). If there is a newer insiders version number, then look for the "asset" (again using the git-hub API) for that release that has our VSIX name in it. Download that VSIX, install it, and prompt the user to reload.

That's basically it.

There is some special handing (also adapted from the Microsoft C/C++ extension) to handle rate limits and HTTP proxy setups.

### DIFF
--- a/package.json
+++ b/package.json
@@ -158,7 +158,24 @@
                                 "sarifFile",
                                 "severityLevel",
                                 "tool"
-                            ]
+                            ],
+                            "enumDescriptions": [
+                                "%automationCat.description%",
+                                "%automationId.description%",
+                                "%baselineState.description%",
+                                "%kind.description%",
+                                "%message.description%",
+                                "%rank.description%",
+                                "%resultFile.description%",
+                                "%resultStartPos.description%",
+                                "%logicalLocation.description%",
+                                "%ruleId.description%",
+                                "%ruleName.description%",
+                                "%runId.description%",
+                                "%sarifFile.description%",
+                                "%severityLevel.description%",
+                                "%tool.description%"
+                            ]        
                         },
                         "ascending": {
                             "type": "boolean"
@@ -194,6 +211,20 @@
                     "description": "%sarif-viewer.openMarkdownPreview.description%",
                     "type": "boolean",
                     "default": false,
+                    "scope": "application"
+                },
+                "sarif-viewer.updateChannel": {
+                    "description": "%sarif-viewer.updateChannel.description%",
+                    "type": "string",
+                    "enum": [
+                        "Default",
+                        "Insiders"
+                    ],
+                    "enumDescriptions": [
+                        "%sarif-viewer.updateChannel.default.description%",
+                        "%sarif-viewer.updateChannel.insiders.description%"
+                    ],
+                    "default": "Default",
                     "scope": "application"
                 }
             }
--- a/package.json
+++ b/package.json
@@ -175,7 +175,7 @@
                                 "%sarifFile.description%",
                                 "%severityLevel.description%",
                                 "%tool.description%"
-                            ]        
+                            ]
                         },
                         "ascending": {
                             "type": "boolean"
--- a/package.json
+++ b/package.json
@@ -257,11 +257,12 @@
         "installNodeModules": "npm install"
     },
     "devDependencies": {
+        "@types/semver": "^7.1.0",
         "@types/glob": "^7.1.1",
         "@types/jquery": "^3.3.30",
         "@types/markdown-it": "0.0.9",
         "@types/mocha": "^7.0.2",
-        "@types/node": "^7.10.7",
+        "@types/node": "^13.13.2",
         "@types/sarif": "^2.1.2",
         "mocha": "^6.2.0",
         "tslint-microsoft-contrib": "^6.0.0",
@@ -272,11 +273,13 @@
         "tslint": "^6.1.1"
     },
     "dependencies": {
+        "semver": "^7.3.2",
         "@microsoft/sarif-multitool": "^2.2.3",
         "json-source-map": "^0.4.0",
         "jsonpointer": "4.0.1",
         "markdown-it": "^9.0.1",
         "requirejs": "^2.3.6",
-        "vscode-nls": "^4.1.2"
+        "vscode-nls": "^4.1.2",
+        "https-proxy-agent": "^5.0.0"
     }
 }
--- a/src/insidersUpdate/gitHubInterfaces.ts
+++ b/src/insidersUpdate/gitHubInterfaces.ts
@@ -1,0 +1,53 @@
+/*!
+ * Copyright (c) Microsoft Corporation. All Rights Reserved.
+ */
+
+/**
+ * For reference, you can find the GitHub API information here:
+ * https://developer.github.com/v3/repos/
+ */
+
+export const GitHubApiBase: string = 'https://api.github.com/repos';
+
+ /**
+  * Contains the minimum information needed to inspect a git-hub release.
+  */
+export interface GitHubRelease {
+    /**
+     * The tag name for the release in the form of v<semver>-insiders.
+     * I.e. v3.2020.424006-insiders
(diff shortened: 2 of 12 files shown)


src/insidersUpdate/updateCheck.ts
```diff
@@ -0,0 +1,301 @@
...
+ * for that release and find the VSIX.
+ * This is done using this GitHub API.
+ * https://developer.github.com/v3/repos/releases/#list-assets-for-a-release
+ * Once we find the asset with the correct "name", it is downloaded and installed.
+ * This implementation was adapted from the Microsoft C/C++ extension.
+ * https://github.com/microsoft/vscode-cpptools/blob/58c50dc38b1a3ebcae8139b28c0904d468d11e6e/Extension/src/githubAPI.ts
+ */
+
+const gitHubRepo: string = 'Microsoft/sarif-vscode-extension';
+const vsixAssetName: string = 'Microsoft.Sarif-Viewer.vsix';
```
jeffersonking, line 35: Broader than this review, but we should align the VSIX file name with our extension ID `ms-sarifvscode.sarif-viewer`.

src/insidersUpdate/updateCheck.ts
```diff
@@ -0,0 +1,301 @@
...
+    const gitHubReleaseResponse: GitHubRelease[] = await downloadOverHttpsAsJsonObject(vscode.Uri.parse(`${GitHubApiBase}/repos/${gitHubRepo}/releases`), 'gitHubReleases.json', userAgentHeaders);
+
+    const releasesToInspect: GitHubRelease[] = gitHubReleaseResponse.slice(0, maxNUmberOfReleasesToInspect);
+    for (const release of releasesToInspect) {
+        if (!release.tag_name.charAt(0).invariantEqual('v')) {
+            continue;
+        }
+
+        const candidateReleaseVersion: semver.SemVer = new semver.SemVer(release.tag_name.substr(1));
+        if (candidateReleaseVersion.prerelease.find((tag) => typeof tag === 'string' && tag.invariantEqual('insiders')) === undefined) {
```
jeffersonking, line 175: No need for `new`. No need to `find`.

```
import {prerelease} from 'semver'
if (!prerelease(release.tag_name.substr(1)).some(tag => typeof tag === 'string' && tag.invariantEqual('insiders')) {
```

src/insidersUpdate/updateCheck.ts
```diff
@@ -0,0 +1,301 @@
...
+            continue;
+        }
+
+        const extensionVersion: semver.SemVer = new semver.SemVer(versionString);
+        if (candidateReleaseVersion.compare(extensionVersion) <= 0) {
```
jeffersonking, line 180: Use `lte` from `semver` instead.

src/insidersUpdate/updateCheck.ts
```diff
@@ -0,0 +1,301 @@
...
+        return false;
+    }
+
+    const vsixUri: vscode.Uri | undefined = await tryDownloadVsix(gitHubAsset);
+    if (!vsixUri) {
+        return false;
+    }
+
+    // Since everything is highly async and it can take a while to download the VSIX, make sure one
+    // more time that the user is on the insiders version.
```
jeffersonking, line 275: This comment is misleading to me. I think the point is: "The download can take a while. The user can change the "update channel" setting during this time. Thus, we need to check one more time."

jeffersonking: If you concerned about that, then if they toggled the switch off/on during this time, it could cause multiple simultaneous downloads/updates correct?

src/insidersUpdate/gitHubInterfaces.ts
```diff
@@ -0,0 +1,93 @@
...
+    /**
+     * The tag name for the release in the form of v<semver>-insiders.
+     * I.e. v3.2020.424006-insiders
+     */
+    readonly tag_name: string;
+
+    /**
+     * The Git Hub release ID.
+     */
+    readonly id: number;
```
jeffersonking, line 30: I would not know without the comment! 😂

package.json
```diff
@@ -158,6 +158,23 @@
                                 "sarifFile",
                                 "severityLevel",
                                 "tool"
+                            ],
+                            "enumDescriptions": [
```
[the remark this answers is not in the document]
GabeDeBacker: Adding the descriptions that I forgot to add  (when doing the localization stuff) that show up in the settings UI for each of the enumerated values.

---
In reply to: [416019288](https://github.com/microsoft/sarif-vscode-extension/pull/267#discussion_r416019288) [](ancestors = 416019288)

src/insidersUpdate/gitHubInterfaces.ts
GabeDeBacker: LOL. Yeah. I know it does get a little out of handle. Our stance is that if it is public (or exported in this case) then it is documented which TSLint is enforcing.

It's just easier to document it than have one off disable TSLint's everywhere.

---
In reply to: [416058245](https://github.com/microsoft/sarif-vscode-extension/pull/267#discussion_r416058245) [](ancestors = 416058245)

src/insidersUpdate/updateCheck.ts
GabeDeBacker: Honestly not sure why the ID isn't Microsoft.SarifViewer.. But yeah, we can discuss changing it althought the VSIX file name isn't really relavent.

---
In reply to: [416030790](https://github.com/microsoft/sarif-vscode-extension/pull/267#discussion_r416030790) [](ancestors = 416030790)

src/insidersUpdate/updateCheck.ts
GabeDeBacker: English is my 3rd language, especially grammar.

---
In reply to: [416055759](https://github.com/microsoft/sarif-vscode-extension/pull/267#discussion_r416055759) [](ancestors = 416055759)

GabeDeBacker: Yep. Correct. I should have handled that. Added code for that.

---
In reply to: [416057404](https://github.com/microsoft/sarif-vscode-extension/pull/267#discussion_r416057404) [](ancestors = 416057404)

src/insidersUpdate/updateCheck.ts
jeffersonking: It's not, but I thought I had the wrong extension for a second.

jeffersonking: Btw, the "Full" Visual Studio Sarif extension downloads as "Microsoft.Sarif.Viewer.vsix".

src/insidersUpdate/gitHubInterfaces.ts
jeffersonking: I feel like rule "public == docs" came from a world with more than just 2 access modifiers.

src/insidersUpdate/updateCheck.ts
```diff
@@ -0,0 +1,329 @@
...
+            continue;
+        }
+
+        const candidateReleaseVersion: semver.SemVer = new semver.SemVer(release.tag_name.substr(1));
+        if (candidateReleaseVersion.prerelease.find((tag) => typeof tag === 'string' && tag.invariantEqual('insiders')) === undefined) {
+            continue;
+        }
+
+        const extensionVersion: semver.SemVer = new semver.SemVer(versionString);
+        if (semver.lte(releaseSemVer, extensionVersion)) {
```
jeffersonking, line 192: This looks like a `string` compared with an `object`. Should be two `string`s.

Issue #261, opened 2020-04-23
Consider adding an option to supress upgrade prompt.


GabeDeBacker:
When users open SARIF files directly, they are constantly getting asked to upgrade SARIF files with either Save (Temp) or Save (As).

We should consider adding an option that allows:
Prompt -> Prompts the user
UpgradeTemp-> Creates a temporary file
UpgradeInPlace -> Upgrades the SARIF in place.

jeffersonking:
If we go this direction, would be good to with common pattern is to allow the user to [Don't Show Again...]. Which goes with Gabe's suggestion of saving to config.

We have a lot of choice buttons, so perhaps the first message should just have the options:
[Upgrade][Don't Show Again]

Note: The "No" option is handled by convention via the [x]. Aka the [No] is an anti-pattern.

Subsequent prompts can handle "save temp" "save as" and also "remember my choice for future...".

PR #268, opened 2020-04-26
Please, no more upgrade prompts :)

GabeDeBacker:
The PR addresses issue #261 
Add a configuration that allows the user to effectively disable the upgrade prompt and have SARIF files that need to be upgraded done so automatically as a temporary file.

@jeffersonking and I have been discussing that there are at least two types of consumers of the SARIF viewer, there are SARIF producers and SARIF consumers. In the case where you are a SARIF producer, seeing the prompt is nice because it gives you and indication that the SARIF producer is out of date. As a consumer, well, you would probably always want to update.

For the time being, we have consumers that will always want an automatic upgrade any out of date SARIF logs,

So the behavior of this change is as follows:
The existing behavior remains where a user is "Prompted" for upgrade. They have options [Upgrade] or [Don't show me this again]. If the users chooses "either" of these options, a temporary file is created. In the case of "Don't show me this again", a configuration is updated to suppress the prompts and always create a temporary file. Clicking the "X" or pressing escape cancels the upgrade\open flow applying no changes. 

The option for "Save As" was removed, because, well, if you want to save the converted file, just use File\Save AS from VSCode.

### DIFF
--- a/src/fileConverter.ts
+++ b/src/fileConverter.ts
@@ -222,7 +222,7 @@ export class FileConverter {
     public static async upgradeSarif(sarifFile: vscode.Uri, sarifVersion: SarifVersion | undefined, sarifSchemaVersion: SarifVersion | undefined, upgradeOptions?: UpgradeSarifOptions): Promise<vscode.Uri | undefined> {
         const desiredUpgradeOption: UpgradeSarifOptions = upgradeOptions ?? vscode.workspace.getConfiguration(Utilities.configSection).get(FileConverter.UpgradePromptSettingName, DefaultUpgradeOption);
 
-        if (desiredUpgradeOption === 'Prompt') {
+        if (desiredUpgradeOption !== 'Always') {
             const upgradeMessage: string = sarifSchemaVersion ?
              localize('converterTool.Upgrade.AskWithSchema', "Sarif schema version {0} is not supported. Upgrade to the latest schema version {1}?", sarifSchemaVersion.original, FileConverter.MultiToolCurrentSchemaVersion.original) :
              localize('converterTool.Upgrade.AskWithVersion', "Sarif version {0} is not supported. Upgrade to the latest version {1}?",
--- a/src/fileConverter.ts
+++ b/src/fileConverter.ts
@@ -42,7 +42,7 @@ export interface UpgradeCheckInformation {
 /**
  * Options used during upgrade of SARIF to later schema version.
  */
-export type UpgradeSarifOptions = 'Prompt' | 'Temporary';
+export type UpgradeSarifOptions = 'Prompt' | 'Always';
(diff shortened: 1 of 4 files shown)


src/fileConverter.ts
```diff
@@ -219,67 +217,40 @@ export class FileConverter {
...
 
-            if (sarifSchemaVersion) {
-                upgradeMessage = localize('converterTool.Upgrade.AskWithSchema', "Sarif schema version {0} is not supported. Upgrade to the latest schema version {1}?", sarifSchemaVersion.original, FileConverter.MultiToolCurrentSchemaVersion.original);
-            } else {
-                upgradeMessage = localize('converterTool.Upgrade.AskWithVersion', "Sarif version {0} is not supported. Upgrade to the latest version {1}?",
-                    sarifVersion && sarifVersion.original ? sarifVersion.original : localize('converterTool.Upgrade.UnknownVersion', "Unknown"), FileConverter.MultiToolCurrentSchemaVersion.original);
-            }
+            const neverShowAgainMessageItem: vscode.MessageItem = {
+                title: localize('converterTool.Upgrade.NeverShow', "Don't Show Again")
+            };
```
sherryyshi, line 237: maybe the options can be "Yes" "Always" and "Never"?

sherryyshi: It's not always obvious that "Don't show again" means no... Although I suppose it is in this case, so you can take or leave that suggestion

---
In reply to: [416045107](https://github.com/microsoft/sarif-vscode-extension/pull/268#discussion_r416045107) [](ancestors = 416045107)

jeffersonking: I do think more clarity is needed. Just for context (and not justification), the we were trying to be more consistent with phrasing elsewhere in VS Code such as here:

Dont Show Again

jeffersonking: So I also interpreted "Don't show again" meant as "Don't upgrade".

I need a moment to develop my opinion on this design.

jeffersonking:
*I would advocate for a non-blocking workflow here.* I think it's a fair departure from previous behavior to automatically write to temp.

In case the user wonders where the old prompt went, we would have a non-blocking toast/message explaining how to "Save As...", how to "Never Upgrade", and the explain the relevance of upgrading. This new non-blocking message can be permanently dismissed with a "Don't show me again."

We can also put a "Open upgraded file" (that was already auto created) as a Code Lens link at the top of the SARIF file. This could take the pressure off of the toast/message. It's the Producer Personas that open these files. They are also the ones more likely to save-as-upgrade. I think we agree the Consumer Personas get a new build frequently, so upgrading is less relevant.

jeffersonking:
Met with @GabeDeBacker. Near-term design will have the choices "Once" and "Always".

I think this is inline with @sherryyshi feedback also.

Issue #275, opened 2020-04-29
The "closeLogs" API doesn't work if a SARIF file is internally upgraded


GabeDeBacker:
When using the "openLogs" API where a SARIF file is automatically updated to the newer schema, the viewer isn't tracking the original name, only the upgraded temp file.

We need a mechanism to track this so the closeLogs API works properly.

PR #276, opened 2020-04-30
API Enhancements

GabeDeBacker:
This addresses issue #275.

When a log-file is upgraded to the latest schema version, the viewer loses track of the original file, so when the closeLogs API is called, it doesn't have anythning to remove from the viewer. I've made the cheapest fix for now which since we create the temp-files in the same location with the same name is the input file, when closeLog is called, I simply ask the viewer to close the temp file as well. It doesn't matter if it actually exists or not because it is only being used as a key.

The second change is to pass a cancellation token into the openLogs API to allow cancellation for a large number of logs files.

The third change is to allow consumers of the API to pass in root paths that they know about in the openLogFiles API as well as the ability to set the directly.

All unit tests passed and verification with another extension calling the API is in progress.

### DIFF
--- a/src/api/sarifViewerApi.ts
+++ b/src/api/sarifViewerApi.ts
@@ -46,7 +46,7 @@ export interface Api {
     closeAllLogs(): Promise<void>;
 
     /**
-     * A set of base URIs to use for mapping remote artifact locations.
+     * A set of URIs bases to use for mapping remote artifact locations.
      */
-    baseUris: ReadonlyArray<vscode.Uri>;
+    uriBases: ReadonlyArray<vscode.Uri>;
 }
--- a/src/api/sarifViewerApi.ts
+++ b/src/api/sarifViewerApi.ts
@@ -20,13 +20,6 @@ export interface OpenLogArguments {
      * The default is true.
      */
     readonly openViewerWhenNoResults?: boolean;
-
-    /**
-     * A set of root paths to use for mapping remote artifact locations.
-     * The paths are used only for this instance of the extension.
-     * They are not persisted in the user settings.
-     */
-    readonly localRootPaths?: vscode.Uri[];
 }
 
 /**
@@ -53,9 +46,7 @@ export interface Api {
     closeAllLogs(): Promise<void>;
 
     /**
-     * A set of root paths to use for mapping remote artifact locations.
-     * The paths are used only for this instance of the extension.
-     * They are not persisted in the user settings.
+     * A set of base URIs to use for mapping remote artifact locations.
      */
-    rootPaths: vscode.Uri[];
+    baseUris: ReadonlyArray<vscode.Uri>;
 }
--- a/src/api/sarifViewerApi.ts
+++ b/src/api/sarifViewerApi.ts
@@ -23,8 +23,8 @@ export interface OpenLogArguments {
 
     /**
      * A set of root paths to use for mapping remote artifact locations.
-     * The paths are used only for this instance
-     * of the extension and are note persisted in the user settings.
+     * The paths are used only for this instance of the extension.
+     * They are not persisted in the user settings.
      */
     readonly localRootPaths?: vscode.Uri[];
 }
@@ -54,8 +54,8 @@ export interface Api {
 
     /**
      * A set of root paths to use for mapping remote artifact locations.
-     * The paths are used only for this instance
-     * of the extension and are note persisted in the user settings.
+     * The paths are used only for this instance of the extension.
+     * They are not persisted in the user settings.
      */
     rootPaths: vscode.Uri[];
 }
--- a/src/api/sarifViewerApi.ts
+++ b/src/api/sarifViewerApi.ts
@@ -20,6 +20,13 @@ export interface OpenLogArguments {
      * The default is true.
      */
     readonly openViewerWhenNoResults?: boolean;
+
+    /**
+     * A set of root paths to use for mapping remote artifact locations.
+     * The paths are used only for this instance
+     * of the extension and are note persisted in the user settings.
+     */
+    readonly localRootPaths?: vscode.Uri[];
 }
 
 /**
@@ -27,11 +34,12 @@ export interface OpenLogArguments {
  */
 export interface Api {
     /**
-     * Opens logs .
+     * Opens logs.
      * @param logs The logs to open.
      * @param openLogArguments Parameters that control how the logs are opened.
+     * @param cancellationToken Token used to cancel the open log request.
      */
-    openLogs(logs: vscode.Uri[], openLogArguments?: OpenLogArguments): Promise<void>;
+    openLogs(logs: vscode.Uri[], openLogArguments?: OpenLogArguments, cancellationToken?: vscode.CancellationToken): Promise<void>;
 
     /**
      * Closes logs.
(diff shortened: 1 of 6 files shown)


src/fileMapper.ts
```diff
@@ -216,61 +233,40 @@ export class FileMapper implements Disposable {
...
-                            } else {
-                                resolvedString = this.rootPaths[rootPathIndex];
-                                input.hide();
-                            }
-                        } else if (this.isLocalFile(validateUri)) {
-                            message = undefined;
-                            resolvedString = validateUri.fsPath;
-                            input.hide();
-                        }
+                const remappedUri: Uri | undefined = this.tryConfigRootPathsUri(uri, uriBase, [validateUri]);
```
roywill, line 255: Do you need to create an array for this?  No override of one?

src/api/sarifViewerApi.ts
```diff
@@ -20,18 +20,26 @@ export interface OpenLogArguments {
      * The default is true.
      */
     readonly openViewerWhenNoResults?: boolean;
+
+    /**
+     * A set of root paths to use for mapping remote artifact locations.
+     * The paths are used only for this instance
+     * of the extension and are note persisted in the user settings.
```
sherryyshi, line 27: >note [](start = 32, length = 4)

not

src/fileMapper.ts
```diff
@@ -342,27 +338,27 @@ export class FileMapper implements Disposable {
      * @param key key used for mapping, if undefined the mapping won't be added if it exists
      */
     private isLocalFile(uri: Uri, key?: string): boolean {
```
roywill, line 340: What do you want to do with . and ..?

src/api/sarifViewerApi.ts
```diff
@@ -43,4 +51,11 @@ export interface Api {
      * Closes all logs.
      */
     closeAllLogs(): Promise<void>;
+
+    /**
+     * A set of root paths to use for mapping remote artifact locations.
+     * The paths are used only for this instance
+     * of the extension and are note persisted in the user settings.
```
sherryyshi, line 58: >note [](start = 32, length = 4)

not

src/extension.ts
```diff
@@ -251,10 +259,24 @@ class SarifExtension implements Api {
...
      */
     public async closeLogs(logs: vscode.Uri[]): Promise<void> {
+        const logsToClose: vscode.Uri[] = [];
+
         for (const log of logs) {
-            this.diagnosticCollection.removeRuns([log]);
+            if (log.isSarifFile()) {
+                logsToClose.push(log);
+                /***
+                 * The temp file path name generated is always results the same output
```
sherryyshi, line 268: >is always results the same outpu [](start = 53, length = 32)

grammar =P

src/utilities.ts
```diff
@@ -177,15 +177,15 @@ export class Utilities {
...
 
     /**
-     * Generates a folder path matching original path in the temp location and returns the path with the file included
+     * Generates a folder path matching original path in the temp location and returns the path with the file included.
+     * The function will always return the same output for the same given input.
      * @param filePath original file path, to recreate in the temp location
-     * @param hashValue optional hash value to add to the path
      */
-    public static generateTempPath(filePath: string, hashValue?: string): string {
+    public static generateTempPath(filePath: string): string {
```
roywill, line 184: Why is this not a URI and return a URI

src/utilities.ts
```diff
@@ -177,15 +177,15 @@ export class Utilities {
...
         const pathObj: path.ParsedPath = path.parse(filePath);
-        let basePath: string = path.join(Utilities.SarifViewerTempDir, hashValue || '');
+        let basePath: string = path.join(Utilities.SarifViewerTempDir, '');
```
roywill, line 186: Why join is this to get a trailing \?

src/utilities.ts
GabeDeBacker: Because there is a lot of other code that still uses strings. The new extension that Jeff is working on will be ALL Uri :)

---
In reply to: [418139508](https://github.com/microsoft/sarif-vscode-extension/pull/276#discussion_r418139508) [](ancestors = 418139508)

src/fileMapper.ts
GabeDeBacker: Good question. :) There should be a lot of path.normalzie in here somewhere... Let me try that out!

---
In reply to: [418137588](https://github.com/microsoft/sarif-vscode-extension/pull/276#discussion_r418137588) [](ancestors = 418137588)

src/fileMapper.ts
```diff
@@ -174,7 +191,7 @@ export class FileMapper implements Disposable {
      * Shows the input box with message for getting the user to select the mapping
      * @param uri uri of the file that needs to be mapped
      */
-    private async openRemappingInputDialog(uri: Uri): Promise<string | undefined> {
+    private async openRemappingInputDialog(uri: Uri, uriBase?: string): Promise<string | undefined> {
```
sherryyshi, line 194: >uriBase [](start = 53, length = 7)

document

src/fileMapper.ts
```diff
@@ -389,20 +385,27 @@ export class FileMapper implements Disposable {
...
-            const dirParts: string[] = dir.split(path.sep);
-            dirParts.push(originPath.base);
-
+    private tryConfigRootPathsUri(originUri: Uri, uriBase?: string, rootPaths?: Uri[]): Uri | undefined {
+        // Parse the remote URI into directory parts.
+        const originPath: path.ParsedPath = path.parse(originUri.fsPath);
+
+        // Remove the "root" of the path. (For example, if the incoming path is e:\, remove it.)
+        // So given an incoming URI such as "e:\foo\bar\xyz.cpp" we are left with "\foo\bar"
+        const originPathWithoutRoot: string = path.join(originPath.dir.replace(originPath.root, ''), originPath.base);
```
sherryyshi, line 395: >originPath.dir.replace(originPath.root, '') [](start = 56, length = 43)

just making sure - does this work as expected on linux/mac?

GabeDeBacker: Yep.

---
In reply to: [418157101](https://github.com/microsoft/sarif-vscode-extension/pull/276#discussion_r418157101) [](ancestors = 418157101)

src/fileMapper.ts
GabeDeBacker: The "normalizing" of paths goes well beyond this PR. started an email thread and I'll talk to Jeff about how this is being handled in the upcoming new code.

---
In reply to: [418149431](https://github.com/microsoft/sarif-vscode-extension/pull/276#discussion_r418149431) [](ancestors = 418149431,418137588)

src/api/sarifViewerApi.ts
```diff
@@ -53,9 +46,7 @@ export interface Api {
     closeAllLogs(): Promise<void>;
 
     /**
-     * A set of root paths to use for mapping remote artifact locations.
-     * The paths are used only for this instance of the extension.
-     * They are not persisted in the user settings.
+     * A set of base URIs to use for mapping remote artifact locations.
      */
-    rootPaths: vscode.Uri[];
+    baseUris: ReadonlyArray<vscode.Uri>;
```
jeffersonking, line 51: I was off in my suggestion. I just checked the SARIF spec and they refer it to as `uriBase` and `uriBases`.

GabeDeBacker: Fixed. Thanks for catching this before I completed the PR :)

---
In reply to: [418274205](https://github.com/microsoft/sarif-vscode-extension/pull/276#discussion_r418274205) [](ancestors = 418274205)

PR #279, opened 2020-05-15
Integrate new UI

jeffersonking:
This brings in the structural rework Gabe and I have been working on. Major parts are:
- A React-based WebView. Increased accessibility and visual consistency relative to VS Code.
- Simplified more-performant messaging between the extension and web-view.
- Less user-intrusive file remapping/rebasing. In general, much lazier resolution for paths.
- Unified webpack build.

Certain features have been consciously not brought forward. Examples are arbitrary file conversion from external formats and certain less-used settings.

Some trailing work I'll do in parallel with this PR:
- Update the readme and changelog.
- Handle rootpaths and other legacy settings correctly.
- Linting and strictness was relaxed. These will come back in the long tail.
- Restoring localization.

Let me know if I missed anything else. Thanks.


index.css
```diff
@@ -0,0 +1,419 @@
+/* Backwards-engineered from VS Code Webview (id="_defaultStyles"). Do not edit.  */
```
GabeDeBacker, line 1: Is it possible to .gitignore this file and have it "pulled" from VSCode's git repository as part of the build pipeline? (We play this trick in other places as well)

package.json
```diff
@@ -1,326 +1,153 @@
...
+			{
+				"command": "sarif.clearState",
+				"title": "SARIF: Clear State"
+			},
+			{
+				"command": "sarif.updateExtension",
+				"title": "SARIF: Update Extension"
+			},
+			{
+				"command": "sarif.apiOpenLogs",
```
GabeDeBacker, line 85: Do these show up in VSCode's command palette?

GabeDeBacker: I assume we don't want them to

---
In reply to: [430574574](https://github.com/microsoft/sarif-vscode-extension/pull/279#discussion_r430574574) [](ancestors = 430574574)

package.json
```diff
@@ -1,326 +1,153 @@
...
+				"title": "SARIF: Test Open Logs via API"
+			},
+			{
+				"command": "sarif.apiCloseLogs",
+				"title": "SARIF: Test Close Logs via API"
```
GabeDeBacker, line 90: >SARIF:  [](start = 14, length = 7)

If you use category property, you shouldn't need to prefix everything with "SARIF:"

index.html
```diff
@@ -0,0 +1,68 @@
...
+					},
+					filtersColumn: {
+						Columns: {
+							'Baseline': false,
+							'Suppression': false,
+							'Rule': false,
+						},
+					},
+				}
+				const state = localStorage.getItem('state')
```
GabeDeBacker, line 54: >localStorage [](start = 18, length = 12)

Since this HTML has script in it with no real context :) can you comment where "localStorage" and "Store" come from?

index.html
```diff
@@ -0,0 +1,68 @@
...
+				const store = new Store(JSON.parse(state) ?? defaultState, true)
+				const file = 'yourSampleSarif.json'
+				const response = await fetch(file)
+				const log = await response.json()
+				log._uri = `file:///Users/username/projects/${file}`
```
GabeDeBacker, line 59: Is this test code? The log file path and the SARIF file seems not quite correct.

GabeDeBacker:
Are there still some things left to implement here? Looks like the update mechanism has been deleted (or perhaps moved somewhere).

src/panel/codicon.css
```diff
@@ -0,0 +1,426 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
```
GabeDeBacker, line 5: Is this also a candidate to retrieve using git in the build pipeline?

src/shared/index.ts
```diff
@@ -0,0 +1,211 @@
...
+import { ArtifactLocation, Log, Location, Region, Result } from 'sarif'
+
+type JsonLocation = { line: number, column: number } // Unused: pos
+type JsonRange = { value: JsonLocation, valueEnd: JsonLocation } // Unused: key, keyEnd
+export type JsonMap = Record<string, JsonRange>
+
+export type ResultId = [string, number, number]
+export type _Region = number | [number, number] | [number, number, number, number]
+
+// Underscored members are ptional in the source files, but required after preprocessing.
```
GabeDeBacker, line 13: >ptional [](start = 27, length = 7)

Time to install a spell checker :)

index.css
jeffersonking: This file is for development purposes (comments now added).

I think these styles are generated dynamically by VS Code. A quick look in their repo did not find this file.

package.json
jeffersonking: Awesome. Updated. Anyway to not have to repeat the category property?

package.json
jeffersonking: Should not be customer facing. Might be useful for us in this next day or two as we verify the API. Removing it (as well some some associated code in `context/index.ts`).

index.html
jeffersonking: Comments added to the top of the file.

index.html
jeffersonking: Yes, it is not production code. Comments added to the top of the file to clarify.

src/panel/codicon.css
jeffersonking: Conceptually possible. However I don't have line-of-sight. Ideally, they would need to be fetched before the bundling process (to make it into the bundle) and I don't know the hook for that. Would need to research.

Alternatively, can drop it in the out folder, but it results in some extra loose files.

src/shared/index.ts
jeffersonking: Installed. :)

jeffersonking:
If any reviewer wants a walkthrough, I'd be happy to do so.

package.nls.json
```diff
@@ -1,112 +0,0 @@
-{
```
sherryyshi, line 1: Why delete this?

resources/icons/call-no-return.svg
```diff
@@ -1,19 +0,0 @@
-<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 16 16">
- <defs>
```
sherryyshi, line 2: Curious, are these images not being used?

tslint.json
```diff
@@ -1,140 +1,24 @@
...
+		"prefer-const": true,
+
+		"file-header": [
+			true, 
+			{
+				"match": "Copyright \\(c\\) Microsoft Corporation\\. All Rights Reserved\\.\\nLicensed under the MIT License\\.",
+				"default" : "Copyright (c) Microsoft Corporation. All Rights Reserved.\nLicensed under the MIT License."
+			}
+		],
+		"indent": [true, "tabs"],
```
sherryyshi, line 17: >"tabs" [](start = 19, length = 6)

boo :P

tslint.json
```diff
@@ -1,140 +1,24 @@
...
-    "linterOptions": {
-        "exclude": [
-            "node_modules/**",
-            "out/**",
-            "src/test/**",
-            "src/explorer/**"
-        ]
-    }
+{
+	"rules": {
```
sherryyshi, line 2: why are there so many changes to this file? seems odd to completely change all the coding styles in one PR

package.nls.json
jeffersonking: There was a high amount of changes/adds/deletes to the strings so the old `package.nls.json` was out of date. I felt it's easier to just create new one (rather than diff it) in this situation. After we've stabilized on functionality, it will be recreated.

src/artifactContentFileSystemProvider.ts
```diff
@@ -1,317 +0,0 @@
-/*!
- * Copyright (c) Microsoft Corporation. All Rights Reserved.
- */
-
-import * as vscode from "vscode";
-import * as sarif from "sarif";
-import * as fs from "fs";
-import * as path from "path";
```
sherryyshi, line 8: Lots of deleted code under here? intentional?

src/context/index.ts
```diff
@@ -0,0 +1,249 @@
...
+			Uri.file('/Users/jeff/projects/sarif-tutorials/samples/3-Beyond-basics/bad-eval-related-locations.sarif'),
+			Uri.file('/Users/jeff/projects/sarif-tutorials/samples/3-Beyond-basics/bad-eval-with-code-flow.sarif'),
+		])
+	})
+	commands.registerCommand('sarif.apiCloseLogs', async () => {
+		const sarifExt = extensions.getExtension('Jeff.sarif-vscode')
+		if (!sarifExt.isActive) await sarifExt.activate()
+		await sarifExt.exports.closeLogs([
+			Uri.file('/Users/jeff/projects/sarif-tutorials/samples/3-Beyond-basics/automation-details.sarif'),
+		])
```
sherryyshi: remove local paths

resources/icons/call-no-return.svg
jeffersonking: These were removed with a feature that "visualiz[ed] codeflow step level changes". Some of the rationale for removal:
* We did not get feedback it was a critical feature.
* The icons have an adverse effect on indentation.
* The icons non-standard and not thematically aligned with VS Code.
* Added complexity to the code flows feature.
* It was an experimental feature we and we had not gotten user feedback on it.

That said, please do take a look at the feature (it's still in the old branch). If you feel it's important, we can bring it back.

tslint.json
jeffersonking: Sorry. :-) I would make the case this is more [accessible](https://www.reddit.com/r/javascript/comments/c8drjo/nobody_talks_about_the_real_reason_to_use_tabs/) but I'm game to adapt to your norms.

src/context/index.ts
```diff
@@ -0,0 +1,249 @@
...
+	const panel = new Panel(context, baser, store)
+	if (uris.length) await panel.show()
+	disposables.push(commands.registerCommand('sarif.showPanel', () => panel.show()))
+
+	// Suggest In-Project Sarif Files
+	;(async () => {
+		const urisSarifInWorkspace = await workspace.findFiles('**/*.sarif', '.sarif/**/*.sarif')
+		const count = urisSarifInWorkspace.length
+		if (!count) return
+		if (await window.showInformationMessage(`Discovered ${count} SARIF logs in your workspace.`, 'View in SARIF Panel')) {
```
sherryyshi, line 97: Add a switch to turn this off? this isn't ideal for very large workspace folders, and a lot of people don't like it when they are flooded by notifications

jeffersonking: In the process of experimenting and design this feature. Agreed on the flooding. It has been turned off.

src/context/Baser.ts
```diff
@@ -0,0 +1,174 @@
...
+	}
+
+	// Hacky.
+	// Note: Uri.parse()
+	// Uri.parse('a/b.c')	 => file:///a/b.c
+	// Uri.parse('/a/b.c')	 => file:///a/b.c
+	// Uri.parse('c:\a\b.c') => c:a%08.c
+	private async pathExists(path: string) {
+		try {
+			await workspace.openTextDocument(Uri.parse(path))
```
sherryyshi, line 63: function name does not match the functionality. Rename to "openIfExists" or something?

tslint.json
jeffersonking: Most of the preexisting lint rules were added relatively recently (last month). Before that it was a yet a different set of rules.

I'm working to add as many as I can back, but some of the rules require non-trivial changes to satisfy. We can decide if we want to hold on delivering functionality for these linting changes.

src/context/Baser.ts
```diff
@@ -0,0 +1,174 @@
...
+					filters: { 'Matching file' : [extension] },
+					// Consider allowing folders.
+				})
+				if (!files?.length) return // User cancelled.
+
+				const partsOld = artifactPath.split('/')
+				const partsNew = files[0]?.path.split('/')
+				if (partsOld.last !== partsNew.last) {
+					void window.showErrorMessage(`File names must match: "${partsOld.last}" and "${partsNew.last}"`)
+					return
```
sherryyshi, line 140: offer to try again?

src/context/Baser.ts
jeffersonking: Admittedly "Hacky" per the comments. We are using `openTextDocument` to test the existence of documents as VS Code does not provide a dedicated existence API. The similar Node `fs` API does not resolve custom URI schemes in the same way that VS Code does otherwise we would use that. I'll update the comments.

src/context/loadLogs.ts
```diff
@@ -0,0 +1,94 @@
...
+				for (const [i, oldLog] of logsToUpgrade.entries()) {
+					progress.report({
+						message: `Upgrading ${i + 1} of ${upgrades} log${upgrades === 1 ? '' : 's'}...`,
+						increment: 1 / upgrades * 100
+					})
+					await new Promise(r => setTimeout(r, 0)) // Await needed for progress to update
+					const {fsPath} = Uri.parse(oldLog._uri)
+					try {
+						const tempPath = upgradeLog(fsPath)
+						const file = fs.readFileSync(tempPath, 'utf8') // Assume scheme file.
```
sherryyshi, line 48: >utf8 [](start = 46, length = 4)

what about binary content?

src/context/Baser.ts
jeffersonking: I think that makes sense.

In the larger picture, the "Locate" dialog (although less intrusive than the previous design) is still too intrusive and noisy. The next design to be trialed is to have the "Locate..." button in the details pane. If that is implemented, then it would be relatively easy to "Locate" again.

src/context/update.ts
```diff
@@ -0,0 +1,47 @@
...
+import { tmpNameSync } from 'tmp'
+import { commands, ProgressLocation, Uri, window } from 'vscode'
+import '../shared/extension'
+
+// Usage:
+// commands.registerCommand('sarif.updateExtension', async () => await update())
+export async function update() {
+	const fileName = tmpNameSync({ postfix: '.vsix' })
+	const stream = fs.createWriteStream(fileName)
+	const url = new URL('https://github.com/jeffersonking/sarif-vscode/releases/download/Latest/sarif-vscode-0.0.0.vsix')
```
sherryyshi, line 15: >https://github.com/jeffersonking/sarif-vscode/releases/download/Latest/sarif-vscode-0.0.0.vsix [](start = 22, length = 94)

?

src/context/loadLogs.ts
jeffersonking: SARIF files are by specification text (JSON) files. We also own the Multitool that outputs the temp file (text) that leads to this code path. Thus it should be reasonable to assume that a binary file will not be encountered here.

src/context/update.ts
```diff
@@ -0,0 +1,47 @@
...
+
+import { https } from 'follow-redirects'
+import fs from 'fs'
+import { tmpNameSync } from 'tmp'
+import { commands, ProgressLocation, Uri, window } from 'vscode'
+import '../shared/extension'
+
+// Usage:
+// commands.registerCommand('sarif.updateExtension', async () => await update())
+export async function update() {
```
sherryyshi, line 12: >update [](start = 22, length = 6)

Are we removing automatic insider updates?

src/context/update.ts
jeffersonking: Custom update functionality that was being used to provide "insiders" updates. `update()` is not current being used, but may become useful to my team (SARIF) in certain cases. We do not share your same Windows pipeline and the release process is thus a bit different.

src/context/update.ts
jeffersonking: Looks like I accidentally answered this question above. If you feel strongly about having this here, I'd be happy to remove it.

src/context/loadLogs.ts
sherryyshi: ahh just read the PR description that support for that will be dropped? I remember there was a bug filed specifically for that, and was resolved by the addition of the support recently. Would need to go and change the status of that bug then...

---
In reply to: [430753208](https://github.com/microsoft/sarif-vscode-extension/pull/279#discussion_r430753208) [](ancestors = 430753208)

package.nls.json
sherryyshi: Read the PR description :) leaving this comment here as a TODO reminder then

---
In reply to: [430734644](https://github.com/microsoft/sarif-vscode-extension/pull/279#discussion_r430734644) [](ancestors = 430734644)

jeffersonking: And I'll do it does Gabe doesn't have to do it again. :-)

package.json
GabeDeBacker: Nope. An extension can have multiple categories. Sorry about that.

tslint.json
GabeDeBacker: I'll be honest Jeff, this is such a dramatic change from any coding style that we have seen that it makes this code a bit difficult to read\understand. A simple example is the lack of semicolons and required bracing for if statements.

Even the original code was more formatted than this syntax.

One of the key reasons that the TSLint configuration was turned up so high was to enforce "consistency" in both style and code architecture in the code with other TypeScript code that Microsoft owns.

VSCode, for example, actually has a git extension installed that requires your code to be formatted a certain way before you can push it even to your own fork\branch :)

We should agree on some compromise here.

---
In reply to: [430750778](https://github.com/microsoft/sarif-vscode-extension/pull/279#discussion_r430750778) [](ancestors = 430750778)

src/shared/index.ts
```diff
@@ -0,0 +1,211 @@
...
+
+	startColumn--
+	if (endColumn) endColumn--
+	if (endLine) endLine--
+	return [
+		startLine,
+		startColumn,
+		endLine ?? startLine,
+		endColumn ?? (startColumn + 1)
+	] as [number, number, number, number]
```
GabeDeBacker, line 169: > as [number, number, number, number] [](start = 2, length = 36)

Isn't this redundant?

jeffersonking: Typescript infers this as `number[]`. We want a tuple of exact four numbers.

install.js
```diff
@@ -0,0 +1,8 @@
+// F5 -> npm start -> npm prestart -> this script.
+// Runs `npm install` only if needed. Otherwise skips (and saves time).
+
+const fs = require('fs')
```
roywill, line 4: Where are all the semicolons.  Make this valid TypeScript

install.js
```diff
@@ -0,0 +1,8 @@
...
+const { execSync } = require('child_process')
```
roywill, line 5: What type is this?

tsconfig.json
```diff
@@ -1,33 +1,20 @@
...
-        "lib": [
-            "dom",
-            "es6"
-        ],
-        "sourceMap": true,
-        "rootDir": "src",
-        /* Strict Type-Checking Option */
-        "strict": true,   /* enable all strict type-checking options */
-        /* Additional Checks */
-        "noUnusedLocals": true, /* Report errors on unused locals. */
```
roywill, line 16: Why are these all turned off?

tsconfig.json
```diff
@@ -1,33 +1,20 @@
...
-        "src/explorer"
-    ]
-}
\ No newline at end of file
+	"compilerOptions": {
+		"module": "commonjs",
+		"target": "es2019",
+		"sourceMap": true,
+		
+		"experimentalDecorators": true,
```
roywill, line 7: >"experimentalDecorators": true, [](start = 2, length = 31)

What does this do?

tslint.json
```diff
@@ -1,140 +1,24 @@
...
-        "trailing-comma": false,
-        "switch-final-break": false,
-        "strict-string-expressions": false,
-        "newline-before-return": false,
-        "object-literal-sort-keys": false,
-        "no-suspicious-comment": false,
-        "export-name": false,
-        "no-magic-numbers": false,
-        "no-unsafe-any": false,
-        "no-unbound-method": false,
```
roywill, line 103: Wow.

install.js
jeffersonking: Not following the questions. The "type" is this: https://nodejs.org/api/child_process.html#child_process_child_process_execsync_command_options. Is that what you meant?

install.js
jeffersonking: Technically valid with or without semicolons and in TypeScript or JavaScript. We change the linter to semi-colons but that would be more easily done in a separate PR.

Note: This script is intentionally not Typescript as it is targeted to run in NodeJS.

tsconfig.json
jeffersonking: Decorators are best explained [here](https://github.com/tc39/proposal-decorators). We mostly use decorators for for Mobx-related modifiers such as `@observable` `@observer` `@action` and `@computed`. However it is useful for other libraries also such as `@autobind`.

tsconfig.json
jeffersonking: In short, we *do* want these enabled. However there's a technical block that I didn't have the time to solve yet.

Attempted explanation: We augment the existing SARIF objects (and thus types) to avoid having two nearly identical models (as the old extension had). Cleanly expressing this in the TypeScript type system (and the in-transition states) will require a pass through the codebase. Just in case there wasn't agreement in this new UI (which doesn't seem to be case now), I didn't want to do potential throw-away work.

jeffersonking: @roywill TypeScript strict is now enabled. I have committed the accompanying edits for conformance.

jeffersonking: @roywill All other flags are now at parity. Note: the diff is not 1-to-1 as `strict` by definition enables some of these other flags. Also note: `checkJs` is no longer `true` as we no longer need to `allowJs`.

src/context/index.ts
```diff
@@ -0,0 +1,232 @@
...
+import '../shared/extension'
+import { Baser } from './Baser'
+import { loadLogs } from './loadLogs'
+import { Panel } from './Panel'
+
+declare module 'vscode' {
+	interface Diagnostic {
+		_result: Result
+	}
+}
```
GabeDeBacker, line 17: Just derive a class from diagnostic, no need to extend the interface. VSCode will hand back an instance of your class.

src/context/index.ts
```diff
@@ -0,0 +1,232 @@
...
+			return new Selection(
+				line.range.start.line,
+				line.firstNonWhitespaceCharacterIndex,
+				line.range.end.line,
+				line.range.end.character,
+			)
+		})()
+	}
+
+export class Store {
```
GabeDeBacker, line 46: Can we put each class in it's own file? We typically do this for each interface and class to keep singe file sizes smaller.

src/context/index.ts
```diff
@@ -0,0 +1,232 @@
...
+import { loadLogs } from './loadLogs'
+import { Panel } from './Panel'
+
+declare module 'vscode' {
+	interface Diagnostic {
+		_result: Result
+	}
+}
+
+export const regionToSelection = (doc: TextDocument, region: _Region | undefined) => {
```
GabeDeBacker, line 19: >export const regionToSelection =  [](start = 0, length = 33)

Why this syntax versus exporting a function?

src/context/index.ts
```diff
@@ -0,0 +1,232 @@
...
+		const count = urisSarifInWorkspace.length
+		if (!count) return
+		if (await window.showInformationMessage(`Discovered ${count} SARIF logs in your workspace.`, 'View in SARIF Panel')) {
+			store.logs.push(...await loadLogs(urisSarifInWorkspace))
+			panel.show()
+		}
+	}) // Disabled while we evaluate the future of this feature.
+
+	// Diagnostics
+	const diagsAll = languages.createDiagnosticCollection('sarif')
```
GabeDeBacker, line 105: >sarif [](start = 56, length = 5)

This should technically be all uppercase since it's an ancronym

src/context/index.ts
```diff
@@ -0,0 +1,232 @@
...
+			})())
+		: (() => {
+			const line = doc.lineAt(region)
+			return new Selection(
+				line.range.start.line,
+				line.firstNonWhitespaceCharacterIndex,
+				line.range.end.line,
+				line.range.end.character,
+			)
+		})()
```
GabeDeBacker, line 43: I believe if statements would be much more readable here. This ternary operator spans ~20 lines.

src/context/index.ts
```diff
@@ -0,0 +1,232 @@
...
+export class Store {
+	static extensionPath: string | undefined
+	static globalState: Memento
+
+	@observable.shallow logs = [] as Log[]
+	@computed public get results() {
+		const runs = this.logs.map(log => log.runs).flat()
+		return runs.map(run => run.results).filter(run => run).flat() as Result[]
+	}
+	@computed public get distinctArtifactNames() {
```
GabeDeBacker, line 55: >public [](start = 11, length = 6)

Curious as to why you use access modifiers here, but not on the constructor. I'd prefer access modifiers everywhere.

src/context/index.ts
```diff
@@ -0,0 +1,232 @@
...
+	Store.globalState = context.globalState
+	disposables.push(commands.registerCommand('sarif.clearState', () => {
+		context.globalState.update('view', undefined)
+		commands.executeCommand('workbench.action.reloadWindow')
+	}))
+	const store = new Store()
+
+	// Boot
+	const uris = await workspace.findFiles('.sarif/**/*.sarif')
+	store.logs.push(...await loadLogs(uris))
```
GabeDeBacker, line 81: We should consider making this an option.

src/context/index.ts
```diff
@@ -0,0 +1,232 @@
...
+
+	// Basing
+	const urisNonSarif = await workspace.findFiles('**/*', '.sarif') // Ignore folders?
+	const fileAndUris = urisNonSarif.map(uri => [uri.path.split('/').pop(), uri.path])  as [string, string][]
```
GabeDeBacker, line 85: >uri [](start = 38, length = 3)

I'm assuming this is a vscode.uri? (can't tell without strong typing). I assume we only work on file-scheme URIs, so this code should be checking for that and using "fspath" not path. Also, use path.parse instead of using "split" because in this case the separator is incorrect for Windows.

src/context/index.ts
```diff
@@ -0,0 +1,232 @@
...
+	const baser = new Baser(mapDistinct(fileAndUris), store)
+
+	// Panel
+	const panel = new Panel(context, baser, store)
+	if (uris.length) await panel.show()
+	disposables.push(commands.registerCommand('sarif.showPanel', () => panel.show()))
```
GabeDeBacker, line 91: >disposables.push( [](start = 1, length = 17)

You can use context.subscriptions for this instead of your own disposables array.

src/context/index.ts
```diff
@@ -0,0 +1,232 @@
...
+	}
+
+export class Store {
+	static extensionPath: string | undefined
+	static globalState: Memento
+
+	@observable.shallow logs = [] as Log[]
+	@computed public get results() {
+		const runs = this.logs.map(log => log.runs).flat()
+		return runs.map(run => run.results).filter(run => run).flat() as Result[]
```
GabeDeBacker, line 53: >as Result[] [](start = 64, length = 11)

Is this "cast" necessary?

src/context/index.ts
```diff
@@ -0,0 +1,232 @@
...
+	return {
+		async openLogs(logs: Uri[]) {
+			store.logs.push(...await loadLogs(logs))
+			if (store.results.length) panel.show()
+		},
+		async closeLogs(logs: Uri[]) {
+			for (const uri of logs) {
+				store.logs.removeWhere(log => log._uri === uri.toString())
+			}
+		},
```
GabeDeBacker, line 220: This should remove all diagnostics, decorations, code-lens, etc. as well.

GabeDeBacker: Same for class all logs.

---
In reply to: [434672138](https://github.com/microsoft/sarif-vscode-extension/pull/279#discussion_r434672138) [](ancestors = 434672138)

src/context/index.ts
```diff
@@ -0,0 +1,232 @@
...
+		async closeAllLogs() {
+			store.logs.splice(0)
```
GabeDeBacker, line 222: Since this should be doing cleanup (diagnostics, etc.) shouldn't this just call closeLogs(store.logs)?

src/context/index.ts
GabeDeBacker, line 85: >  as [string, string][] [](start = 83, length = 23)

There seem to be a lot of "as TYPE" statements, are they necessary?

src/context/index.ts
```diff
@@ -0,0 +1,232 @@
...
+		if (!count) return
+		if (await window.showInformationMessage(`Discovered ${count} SARIF logs in your workspace.`, 'View in SARIF Panel')) {
+			store.logs.push(...await loadLogs(urisSarifInWorkspace))
+			panel.show()
+		}
+	}) // Disabled while we evaluate the future of this feature.
+
+	// Diagnostics
+	const diagsAll = languages.createDiagnosticCollection('sarif')
+	const setDiags = (doc: TextDocument) => {
```
GabeDeBacker, line 106: >setDiags [](start = 7, length = 8)

I would prefer these functions be declared outside of this function so readers do not have to switch context from reading the activate function. The reader will lose the code flow of activate,

.vscode/tasks.json
```diff
@@ -1,49 +1,28 @@
...
-            }
-        }
-    ]
-}
\ No newline at end of file
+	"version": "2.0.0",
+	"tasks": [
+		{
+			"type": "npm",
+			"script": "start",
```
GabeDeBacker, line 6: >"script": "start", [](start = 3, length = 18)

Have the script depend on an "npm install" task so you can clone the repo and just press F5. It will also cause npm install if a developer does a pull as well (added benefit).

index.html
```diff
@@ -0,0 +1,72 @@
...
+<html lang="en">
+	<head>
+		<link rel="stylesheet" href="index.css">
+		<style>
+			code { font-family: SF Mono, Menlo, Monaco, 'Courier New', monospace; }
+		</style>
+	</head>
+	<body class="vscode-dark">
+		<div id="root" style="margin: 50px 300px; position: relative;"></div><!-- relative for Popover -->
+		<script>
```
GabeDeBacker, line 15: Can the scripts here be done in TypeScript or in another javascript file? It would be nice to have a nice separation between HTML and Script.

install.js
```diff
@@ -0,0 +1,8 @@
...
+
+if (!fs.existsSync('node_modules'))
```
GabeDeBacker, line 7: This can be done in tasks.json, this JS file shouldn't be needed.

package.json
```diff
@@ -1,326 +1,145 @@
...
+				"extensions": [
+					".sarif"
+				]
+			}
+		],
+		"commands": [
+			{
+				"command": "sarif.showPanel",
+				"category": "SARIF",
+				"title": "Show Panel"
```
GabeDeBacker, line 75: >"Show Panel [](start = 13, length = 11)

You're really going to want to start with localization already set up. Adding it later is a pain.

src/context/Baser.spec.ts
```diff
@@ -0,0 +1,96 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+import assert from 'assert'
```
GabeDeBacker, line 4: >port assert [](start = 2, length = 11)

Are .spec files tests? Is it possible to put them in a different directory?

src/context/Baser.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+import { mockVscode } from '../test/mockVscode' // Must come before Baser.
```
GabeDeBacker, line 5: >// Must come before Baser. [](start = 48, length = 26)

Why?

sherryyshi:
> 

Would be good to have documentation since this will be publicly distributed, even if the file names may seem obvious

---
Refers to: src/context/index.d.ts:12 in 3f9eaf8. [](commit_id = 3f9eaf80cba86de95ea3e3467a0e4fcd2b0e89fa, deletion_comment = False)

install.js
jeffersonking: Couldn't find a way to do it in only `tasks.json`. Would appreciate any tips here.

src/context/Baser.spec.ts
jeffersonking: `spec` files are tests. This is Mocha's convention, not mine.

By default I keep unit tests (and other related files such as CSS) next to the file it is testing. High-level tests would go in a different directory. I've seen both conventions at Microsoft. For example here at [Office Fabric aka FluentUI](https://github.com/microsoft/fluentui/tree/master/packages/experiments/src/components/Button). Let's discuss.

src/context/Panel.ts
```diff
@@ -0,0 +1,153 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
```
sherryyshi, line 2: nit: different file name casing on this and a few other files. intentional?

package.json
jeffersonking: I have been consciously keeping tracking the timing balance of this. There were a large amount of strings added and deleted during initial construction and it was faster for me not to synchronize a string table simultaneously.

That said, I think we are in the sweet spot for transitioning right now. Also, I take responsibility for it. I know you already did it once. :-)

index.html
jeffersonking: For Typescript, this would require a separate webpack export (aka "pipeline") so I was hesitant.

We can have the Javascript in a separate file, that's no problem.

.vscode/tasks.json
jeffersonking: The initial motivation was to conditionally not spend time doing `npm install` if `node_modules` was already present to save some dev loop time. We discussed offline this is no big deal, which removes the need for the script, and that should satisfy this comment.

src/context/Panel.ts
jeffersonking: Classes in upper. Everything else in lower. Minor exceptions.

Your team prefers all lower, is that right?

src/context/index.ts
jeffersonking: I use `as` only when necessary. In this case it is to override the default inference of `string[]`.

jeffersonking: I use `as` only when necessary. In this case it is to override the default inference of `string[][]` and increase the specificity.

src/context/Panel.ts
sherryyshi: yeah that's what we've been doing since almost everything is encapsulated in a class =P

---
In reply to: [434751991](https://github.com/microsoft/sarif-vscode-extension/pull/279#discussion_r434751991) [](ancestors = 434751991)

src/context/index.ts
jeffersonking: I normally omit `public` since it is the default. So this `public` here (and one below) is a mistake.

src/context/index.ts
jeffersonking: Agreed. This one grew. Will fix.

src/context/index.ts
jeffersonking: Yes.

src/context/index.ts
jeffersonking: I nearly never want the `this` semantic of `function`. So I stick with `arrow functions`. This is true even when the `this` semantic isn't significant just to keep simple.

You're welcome to criticize me for `function activate` a few lines below which is a carryover from the sample code which this all started from. I'm not being self-consistent here and I should. :-)

jeffersonking:
Please give me a chance and some time to address the style-related issues. I'll address the patterns and don't want to bother folks with "proofreading".

If there are any functional or behavioral issues, it would be great to get the spotlight on those. I'd like to see them earlier so I can budget enough time to address them.

src/context/Baser.ts
```diff
@@ -0,0 +1,175 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+import { Uri, window, workspace } from 'vscode'
+import { Store } from '.'
+import '../shared/extension'
+
+export class Baser {
```
GabeDeBacker, line 8: >Baser [](start = 13, length = 5)

Can we rename this to at least "PathBaser" and add a description as to what this code is for?

src/shared/extension.ts
```diff
@@ -0,0 +1,75 @@
...
+		if (typeof aa === 'number' && typeof bb === 'number') return invert * (aa - bb)
+		return 0
+	})
+	return this
+}
+
+!String.prototype.hasOwnProperty('file') &&
+Object.defineProperty(String.prototype, 'file', {
+	get: function() {
+		return this.substring(this.lastIndexOf('/') + 1, this.length)
```
sherryyshi, line 66: >/ [](start = 42, length = 1)

does this work on Windows?

jeffersonking: Need to fix. Thanks for the catch. There is some additional windows testing we (likely just Mukul and I) need to do. It was difficult earlier as I am on a Mac.

src/shared/index.ts
sherryyshi: perhaps make an interface for this

---
In reply to: [431331282](https://github.com/microsoft/sarif-vscode-extension/pull/279#discussion_r431331282) [](ancestors = 431331282)

src/shared/index.ts
```diff
@@ -0,0 +1,214 @@
...
+	},
+} as Record<string, Record<string, boolean>>
+
+export const filtersColumn = {
+	Columns: {
+		'Baseline': false,
+		'Suppression': false,
+		'Rule': false,
+	},
+} as Record<string, Record<string, boolean>>
```
sherryyshi, line 214: >as Record<string, Record<string, boolean>> [](start = 2, length = 42)

Instead of having `as TYPE` would specifying the type after declaration achieve the same thing? If so that's the convention we've been using, and we would prefer that

src/context/index.ts
jeffersonking: Good call.

jeffersonking: Fixed.

src/context/index.ts
jeffersonking: Fixed. Will continue to break part multi-class files as I work.

src/context/index.ts
jeffersonking: Fixed.

src/context/index.ts
jeffersonking: Fixed.

src/context/index.ts
jeffersonking: Fixed.

src/context/Baser.spec.ts
jeffersonking: In transparency: I do like having the unit tests next to the associated source files. 1) It helps me keep them in sync. 2) this pattern is already being used (as mentioned) with CSS.

src/context/index.ts
jeffersonking: This is a proposed future. Since we're going into production, I just removed it from the source.

src/context/index.ts
jeffersonking: Will cleanup.

src/context/index.ts
jeffersonking: This is really syntactic-sugar/explanatory-variable for my own brain. "Disposables" seems to what most people call this. Not sure why VS Code thinks of these as "subscriptions" which means something different in my head.

src/context/index.ts
jeffersonking: Typescript normally infers this as `(Result | undefined)[]. It does not understand `filter(run => run)` removes the undefined values. Thus overriding the type to `Result[]`.

src/context/index.ts
jeffersonking: Diagnostics removals are handled here: [`observe(store.logs, () => workspace.textDocuments.forEach(setDiags))`](https://github.com/microsoft/sarif-vscode-extension/blob/1d83faaaf100aca58b834afe69c3fe471875d104/src/context/index.ts#L68)

Decorations have been fixed and are now handled [here](https://github.com/microsoft/sarif-vscode-extension/blob/1d83faaaf100aca58b834afe69c3fe471875d104/src/context/index.ts#L133)

We currently don't have any code lenses. I think that covers all types.

src/context/Baser.spec.ts
jeffersonking: `mockVscode` mocks/intercepts some other modules such as `vscode`.

I'd ask for some leeway for all test-related code. It's work-in-progress as well as strategy-in-progress. Experimenting with a few different approaches. Discussion is welcome, but ideally at the higher level.

src/context/index.ts
```diff
@@ -0,0 +1,187 @@
...
+			.map(result => new ResultDiagnostic(
+				regionToSelection(doc, result._region),
+				result._message ?? '—',
+				severities[result.level ?? ''] ?? DiagnosticSeverity.Information, // note, none, undefined.
+				result,
+			))
+		diagsAll.set(doc.uri, diags)
+	}
+	workspace.textDocuments.forEach(setDiags)
+	workspace.onDidOpenTextDocument(setDiags)
```
GabeDeBacker, line 66: >onDidOpenTextDocument [](start = 11, length = 21)

Every event registration is disposable

src/context/index.ts
```diff
@@ -0,0 +1,187 @@
...
+	const syncActiveLog = async (doc: TextDocument) => {
+		if (!doc.fileName.match(/\.sarif$/i)) return
+		if (store.logs.some(log => log._uri === doc.uri.toString())) return
+		store.logs.push(...await loadLogs([doc.uri]))
+		panel.show()
+	}
+	workspace.textDocuments.forEach(syncActiveLog)
+	workspace.onDidOpenTextDocument(syncActiveLog)
+	workspace.onDidCloseTextDocument(doc => {
+		if (!doc.fileName.match(/\.sarif$/i)) return
```
GabeDeBacker, line 80: >doc [](start = 7, length = 3)

You can use the doc language id to look for the language defined in package.json

src/context/index.ts
```diff
@@ -0,0 +1,187 @@
...
+
+			const diagnostic = context.diagnostics[0] as ResultDiagnostic | undefined
+			if (!diagnostic) return
+
+			const result = diagnostic?.result
+			panel.select(result)
+
+			const editor = window.visibleTextEditors.find(editor => editor.document === doc)
+			if (!editor) return // When would editor be undef?
+			if (!result) return // Don't clear the decorations until the next result is selected.
```
GabeDeBacker, line 105: >	if (!result) return //  [](start = 2, length = 24)

Move this up.

src/context/index.ts
```diff
@@ -0,0 +1,187 @@
...
+	})
+
+	// Virtual Documents
+	workspace.registerTextDocumentContentProvider('sarif', {
+		provideTextDocumentContent: (uri, token) => {
+			const [logUriEncoded, runIndex, artifactIndex] = uri.path.split('/')
+			const logUri = decodeURIComponent(logUriEncoded)
+			const artifact = store.logs.find(log => log._uri === logUri)?.runs[+runIndex]?.artifacts?.[+artifactIndex]
+			const contents = artifact?.contents
+			if (contents?.text) return contents?.text
```
GabeDeBacker, line 149: >contents?.text [](start = 7, length = 14)

This can have markdown as well.

src/context/index.ts
jeffersonking: The cleanup is accounted for with `store.logs.splice(0)`.

The logic that manages Diagnostics and Decorates watches `store.logs` and will react accordingly.

src/extension/store.ts
```diff
@@ -0,0 +1,31 @@
...
+	}
+	@computed get distinctArtifactNames() {
+		const fileAndUris = this.logs.map(log => [...log._distinct.entries()]).flat()
+		return mapDistinct(fileAndUris)
+	}
+
+	constructor() {
+		intercept(this.logs, objChange => {
+			const change = objChange as unknown as IArrayWillSplice<Log>
+			change.added = change.added.filter(log => this.logs.every(existing => existing._uri !== log._uri))
```
GabeDeBacker, line 27: >change.added = change.added.filter(log => this.logs.every(existing => existing._uri !== log._uri)) [](start = 3, length = 98)

This raises an interesting point which is if a consumer using the APIs gives a log URI to the SARIF extension and then it is modified on the file-system, should the SARIF extension handle this or should the caller of the API handle this.

src/extension/panel.ts
```diff
@@ -0,0 +1,154 @@
...
+			if (!this.panel.active) this.panel.reveal(undefined, true)
+			return
+		}
+
+		const {context, basing, store} = this
+		const {webview} = this.panel = window.createWebviewPanel(
+			'Index', `${this.title}s`, { preserveFocus: true, viewColumn: ViewColumn.Two }, // ViewColumn.Besides steals focus regardless of preserveFocus.
+			{
+				enableScripts: true,
+				localResourceRoots: [Uri.file('/'), ...'abcdefghijklmnopqrstuvwxyz'.split('').map(c => Uri.file(`${c}:`))],
```
GabeDeBacker, line 44: >	localResourceRoots: [Uri.file('/'), ...'abcdefghijklmnopqrstuvwxyz'.split('').map(c => Uri.file(`${c}:`))], [](start = 3, length = 108)

This won't work for UNC shares..

src/extension/panel.ts
```diff
@@ -0,0 +1,154 @@
...
+				const log = store.logs.find(log => log._uri === logUri)
+				const result = store.logs.find(log => log._uri === logUri)?.runs[runIndex]?.results?.[resultIndex]
+				if (!log || !result) return
+				await this.selectLocal(logUri, log._uriUpgraded ?? log._uri, result._logRegion)
+			}
+			if (command === 'setState') {
+				const oldState = Store.globalState.get('view', defaultState)
+				const {state} = message
+				await Store.globalState.update('view', Object.assign(oldState, JSON.parse(state)))
+			}
```
GabeDeBacker, line 116: Throw if command isn't handled???

src/shared/extension.ts
```diff
@@ -0,0 +1,75 @@
...
+	}
+})
+
+!String.prototype.hasOwnProperty('path') &&
+Object.defineProperty(String.prototype, 'path', {
+	get: function() {
+		return this.substring(0, this.lastIndexOf('/')).replace(/^\//g, '')
```
GabeDeBacker, line 73: Isn't this just path.dirname?

src/shared/extension.ts
GabeDeBacker: I think this is just path.basename

---
In reply to: [434777721](https://github.com/microsoft/sarif-vscode-extension/pull/279#discussion_r434777721) [](ancestors = 434777721)

src/shared/extension.ts
```diff
@@ -0,0 +1,75 @@
...
+	interface Array<T> {
+		last: T
+		replace(items: T[]): void // From Mobx, but not showing up.
+		remove(item: T): boolean // From Mobx, but not showing up.
+		removeWhere(predicate: (item: T) => boolean): T | false
+		sortBy<T>(this: T[], selector: Selector<T>, descending?: boolean): Array<T> // Not a copy
+	}
+	interface String {
+		file: string
+		path: string
```
GabeDeBacker, line 23: This bothers me a bit since we should probably be treating everything as URIs, so I could see these extensions on VSCode's URI class.

I know there are places were we have to treat them as strings, but those should be limited to the UI.

If these are only for UI display, then calling them something like fileDisplayName/pathDisplayName would make it clear these strings shouldn't be used for URI manipulation.

jeffersonking: The issue is that we don't have access to the Uri class from within the WebView. Yes, ideally it would be extended from the Uri class. Any ideas for workarounds?

src/extension/regionToSelection.ts
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+import { Selection, TextDocument } from 'vscode'
+import { _Region } from '../shared'
+import '../shared/extension'
+
+export const regionToSelection = (doc: TextDocument, region: _Region | undefined) => {
```
GabeDeBacker, line 8: >const regionToSelection =  [](start = 7, length = 26)

Can this just be a function?

jeffersonking: Changed it to a function.

---
In reply to: [435918721](https://github.com/microsoft/sarif-vscode-extension/pull/279#discussion_r435918721) [](ancestors = 435918721)

src/shared/extension.ts
jeffersonking: No access to vscode.Uri and also any of node's URI or path libraries from within a WebView. When you mention "path.dirname", which API or library are you referring to?

---
In reply to: [435598862](https://github.com/microsoft/sarif-vscode-extension/pull/279#discussion_r435598862) [](ancestors = 435598862)

src/shared/extension.ts
jeffersonking: See comments for "path.dirname".

---
In reply to: [435599059](https://github.com/microsoft/sarif-vscode-extension/pull/279#discussion_r435599059) [](ancestors = 435599059,434777721)

jeffersonking: This will be addressed.

---
In reply to: [434773253](https://github.com/microsoft/sarif-vscode-extension/pull/279#discussion_r434773253) [](ancestors = 434773253)

src/shared/extension.ts
jeffersonking: The issue is that we don't have access to the Uri class from within the WebView. Yes, ideally it would be extended from the Uri class. Any ideas for workarounds?

---
In reply to: [435600491](https://github.com/microsoft/sarif-vscode-extension/pull/279#discussion_r435600491) [](ancestors = 435600491)

src/context/loadLogs.ts
jeffersonking: Binary content is still supported.

If you are referring to "file conversion from external formats". This means conversion from something like AndroidStudio.xml into the SARIF format. For more info, that old feature was here: [https://github.com/microsoft/sarif-vscode-extension/blob/master/src/fileConverter.ts](https://github.com/microsoft/sarif-vscode-extension/blob/master/src/fileConverter.ts)

---
In reply to: [430755897](https://github.com/microsoft/sarif-vscode-extension/pull/279#discussion_r430755897) [](ancestors = 430755897,430753208)

src/shared/index.ts
jeffersonking: Will do.

---
In reply to: [434794240](https://github.com/microsoft/sarif-vscode-extension/pull/279#discussion_r434794240) [](ancestors = 434794240,431331282)

tslint.json
jeffersonking: Sorry. :-) I would make the case this is more accessible but I'm game to adapt to your norms.

---
In reply to: [430737144](https://github.com/microsoft/sarif-vscode-extension/pull/279#discussion_r430737144) [](ancestors = 430737144)

src/panel/codicon.css
jeffersonking: Conceptually possible. However I don't have line-of-sight. Ideally, they would need to be fetched before the bundling process (to make it into the bundle) and I don't know the hook for that. Would need to research.

Alternatively, can drop it in the out folder, but it results in some extra loose files.

---
In reply to: [430586831](https://github.com/microsoft/sarif-vscode-extension/pull/279#discussion_r430586831) [](ancestors = 430586831)

src/context/Baser.ts
jeffersonking: Admittedly "Hacky" per the comments. We are using openTextDocument to test the existence of documents as VS Code does not provide a dedicated existence API. The similar Node fs API does not resolve custom URI schemes in the same way that VS Code does otherwise we would use that. I'll update the comments.

---
In reply to: [430750240](https://github.com/microsoft/sarif-vscode-extension/pull/279#discussion_r430750240) [](ancestors = 430750240)

src/context/Baser.ts
jeffersonking: I think that makes sense.

In the larger picture, the "Locate" dialog (although less intrusive than the previous design) is still too intrusive and noisy. The next design to be trialed is to have the "Locate..." button in the details pane. If that is implemented, then it would be relatively easy to "Locate" again.

---
In reply to: [430751162](https://github.com/microsoft/sarif-vscode-extension/pull/279#discussion_r430751162) [](ancestors = 430751162)

src/context/index.ts
jeffersonking: Typescript thinks it is (Result | undefined)[]. It does not realize the undefined values have been filtered out. Thus the cast.

---
In reply to: [434591947](https://github.com/microsoft/sarif-vscode-extension/pull/279#discussion_r434591947) [](ancestors = 434591947)

index.html
jeffersonking: For Typescript, this would require a separate webpack export (aka "pipeline") so I was hesitant.

We can have the Javascript in a separate file, that's no problem.

---
In reply to: [434716503](https://github.com/microsoft/sarif-vscode-extension/pull/279#discussion_r434716503) [](ancestors = 434716503)

src/context/index.ts
jeffersonking: Line 16: const disposables = context.subscriptions

This is really syntactic-sugar/an-explanatory-variable for my own brain. "Disposables" seems to what most people call this pattern. Not sure why VS Code calls these as "subscriptions" which means something different in my head.

---
In reply to: [434584876](https://github.com/microsoft/sarif-vscode-extension/pull/279#discussion_r434584876) [](ancestors = 434584876)

src/extension/store.ts
jeffersonking: I added a comment to the API to roughly capture this topic.

---
In reply to: [435538028](https://github.com/microsoft/sarif-vscode-extension/pull/279#discussion_r435538028) [](ancestors = 435538028)

index.html
jeffersonking: Moved the bulk of it into "index.js". Note, I left the first script block in as having a separate script file for 9 lines of code seemed overkill. Also note, I couldn't merge it with the other script due to execution order issues.

---
In reply to: [436290240](https://github.com/microsoft/sarif-vscode-extension/pull/279#discussion_r436290240) [](ancestors = 436290240,434716503)

src/context/index.ts
jeffersonking: Is this high priority for you guys?

---
In reply to: [435428710](https://github.com/microsoft/sarif-vscode-extension/pull/279#discussion_r435428710) [](ancestors = 435428710)

src/context/index.ts
jeffersonking: Tracking the general issue of disposables with the other comment.

---
In reply to: [436290264](https://github.com/microsoft/sarif-vscode-extension/pull/279#discussion_r436290264) [](ancestors = 436290264,434584876)

src/shared/index.ts
jeffersonking: Typescript infers this as number[]. We want a tuple of exactly four numbers.

---
In reply to: [431219212](https://github.com/microsoft/sarif-vscode-extension/pull/279#discussion_r431219212) [](ancestors = 431219212)

jeffersonking: Added more explicit types for this.

---
In reply to: [436197296](https://github.com/microsoft/sarif-vscode-extension/pull/279#discussion_r436197296) [](ancestors = 436197296,434794240,431331282)

tslint.json
jeffersonking: Merging threads. Moving Sherry's comment here: "why are there so many changes to this file? seems odd to completely change all the coding styles in one PR".

---
In reply to: [431194797](https://github.com/microsoft/sarif-vscode-extension/pull/279#discussion_r431194797) [](ancestors = 431194797,430750778)

src/panel/codicon.css
jeffersonking: Good news: They just released these icons on NPM. Bad news: There is a blocking bug. I've filed an issue with them to track: [https://github.com/microsoft/vscode-codicons/issues/15](https://github.com/microsoft/vscode-codicons/issues/15). When they fix it, I'll complete this item.

---
In reply to: [436289721](https://github.com/microsoft/sarif-vscode-extension/pull/279#discussion_r436289721) [](ancestors = 436289721,430586831)

tslint.json
jeffersonking: Going track this issue with Gabe's comment about general style issues above. Thus closing this one.

---
In reply to: [436289686](https://github.com/microsoft/sarif-vscode-extension/pull/279#discussion_r436289686) [](ancestors = 436289686,430737144)

src/panel/codicon.css
jeffersonking: They fixed the package. We are no longer checking these files in. Resolved.

---
In reply to: [436327157](https://github.com/microsoft/sarif-vscode-extension/pull/279#discussion_r436327157) [](ancestors = 436327157,436289721,430586831)

src/context/index.ts
jeffersonking: The doc.languageId is currently just "json". I guess we can defined a new language id 'sarif'. But given we only have 1 file extension (and not all JSON files are SARIF files), and we don't have any special language features (above and beyond JSON), is there any other benefit to going this direction?

---
In reply to: [435424162](https://github.com/microsoft/sarif-vscode-extension/pull/279#discussion_r435424162) [](ancestors = 435424162)

src/shared/index.ts
```diff
@@ -0,0 +1,216 @@
...
+	= number // single line
+	| _RegionBytes
+	| _RegionStartEndLineCol
+
+// Underscored members are optional in the source files, but required after preprocessing.
+declare module 'sarif' {
+	interface Log {
+		_uri: string
+		_uriUpgraded?: string
+		_jsonMap?: JsonMap
```
GabeDeBacker, line 23: >? [](start = 10, length = 1)

Per your comment above, this shouldn't be optional?

src/shared/index.ts
```diff
@@ -0,0 +1,216 @@
...
+		_augmented?: boolean
+		_distinct: Map<string, string> // Technically per Run, practically doesn't matter right now.
+	}
```
GabeDeBacker, line 26: So just a thought, I'm curious as to why you wouldn't create another object literal inside the Log interface (and the others as well) for the parsed results.

Something like.

interface Log {
  parsedResults?: {
      uri.: string,
     ...
  }
}

I think this would make the code easier to read when reading the code and trying to remember what "_" is really supposed to mean :)
That one comment will be super easy to miss.

Thoughts?

src/shared/index.ts
jeffersonking: Comment out of date after the recent enablement of Typescript Strict. Will update.

---
In reply to: [437555828](https://github.com/microsoft/sarif-vscode-extension/pull/279#discussion_r437555828) [](ancestors = 437555828)

src/shared/index.ts
jeffersonking: I assume you mean 'parsed_______' here? Or do you really mean Sarif.Result (I don't see why)?

Given the most recent change to Typescript Strict, the optionality will *not* be "time-dependent" (aka before or after processing/augmentation). Now fields can be truly non-optional (or truly optional). And in this evolved world, I would argue we could drop the underscores (aside from a few naming collisions). In practice I don't think there's anywhere we behave differently (or need to know) if the property is processed/augmented/parsed/etc or not. The distinction might not be required.

---
In reply to: [437557263](https://github.com/microsoft/sarif-vscode-extension/pull/279#discussion_r437557263) [](ancestors = 437557263)

src/shared/index.ts
```diff
@@ -0,0 +1,216 @@
...
+	const uri = anyArtLoc.uri ?? runArtLoc?.uri // If index (§3.4.5) is absent, uri SHALL be present.
+	const uriContents = runArtCon?.text || runArtCon?.binary
+			? encodeURI(`sarif:${encodeURIComponent(result._log._uri)}/${result._run._index}/${anyArtLoc.index}/${uri?.file ?? 'Untitled'}`)
+			: undefined
+	return [uri, uriContents]
+}
+
+export const filtersRow = {
+	Level: {
+		'Error': true,
```
GabeDeBacker, line 193: >true [](start = 11, length = 4)

What do the true\false mean here? (and below)?

Can we create a type to help me understand this?

GabeDeBacker: In our repo code, we actually shy away from using bools at all and use a type that is a union of string literals for code readibility.

---
In reply to: [437693237](https://github.com/microsoft/sarif-vscode-extension/pull/279#discussion_r437693237) [](ancestors = 437693237)

jeffersonking: Show/hide, visible/non-visible. I can change. Any preference on the terms used?  There are downstream effects so this may take a minute.

---
In reply to: [437696855](https://github.com/microsoft/sarif-vscode-extension/pull/279#discussion_r437696855) [](ancestors = 437696855,437693237)

src/shared/index.ts
jeffersonking: Changed. I have been pretty consistent with the "as TYPE"-style. I would need time to change every instance over to the ": TYPE" style.

---
In reply to: [434795530](https://github.com/microsoft/sarif-vscode-extension/pull/279#discussion_r434795530) [](ancestors = 434795530)

src/shared/index.ts
jeffersonking: Changed.

---
In reply to: [437704442](https://github.com/microsoft/sarif-vscode-extension/pull/279#discussion_r437704442) [](ancestors = 437704442,437696855,437693237)

src/shared/index.ts
jeffersonking: "By Design" for now. Feel free to re-open if you disagree.

---
In reply to: [437659428](https://github.com/microsoft/sarif-vscode-extension/pull/279#discussion_r437659428) [](ancestors = 437659428,437557263)

src/shared/index.ts
jeffersonking: Updated the comment. Also added a bit more comments for some members.

---
In reply to: [437644402](https://github.com/microsoft/sarif-vscode-extension/pull/279#discussion_r437644402) [](ancestors = 437644402,437555828)

package.nls.json
jeffersonking: Tracking this as an issue ([https://github.com/microsoft/sarif-vscode-extension/issues/281](https://github.com/microsoft/sarif-vscode-extension/issues/281)) and resolving this thread.

---
In reply to: [430756276](https://github.com/microsoft/sarif-vscode-extension/pull/279#discussion_r430756276) [](ancestors = 430756276,430734644)

src/context/update.ts
jeffersonking: It is back. Just pending some proxy work and real life testing.

---
In reply to: [430754296](https://github.com/microsoft/sarif-vscode-extension/pull/279#discussion_r430754296) [](ancestors = 430754296)

jeffersonking:
The update mechanism is back. Just pending some proxy work and real life testing.

---
In reply to: [634160562](https://github.com/microsoft/sarif-vscode-extension/pull/279#issuecomment-634160562) [](ancestors = 634160562)

tslint.json
jeffersonking: Upgraded to the non-deprecated `typescript-eslint` and enabled the `recommended` ruleset with many fewer exceptions than before this PR.

---
In reply to: [432157363](https://github.com/microsoft/sarif-vscode-extension/pull/279#discussion_r432157363) [](ancestors = 432157363)

src/shared/extension.ts
jeffersonking: Unfortunately not.

---
In reply to: [438342421](https://github.com/microsoft/sarif-vscode-extension/pull/279#discussion_r438342421) [](ancestors = 438342421,436180678,435599059,434777721)

jeffersonking:
Due to the number of revisions and comments, the CodeFlow performance is suffering. Thus starting a fresh PR and the review will continue there. See everyone on the other side.

PR #282, opened 2020-06-16
Integrate new UI (Continued)

jeffersonking:
[Note same as previous PR.]

This brings in the structural rework Gabe and I have been working on. Major parts are:
- A React-based WebView. Increased accessibility and visual consistency relative to VS Code.
- Simplified more-performant messaging between the extension and web-view.
- Less user-intrusive file remapping/rebasing. In general, much lazier resolution for paths.
- Unified webpack build.

Certain features have been consciously not brought forward. Examples are arbitrary file conversion from external formats and certain less-used settings.

Some trailing work I'll do in parallel with this PR:
- Update the readme and changelog.
- Handle rootpaths and other legacy settings correctly.
- Linting and strictness was relaxed. These will come back in the long tail.
- Restoring localization.

Let me know if I missed anything else. Thanks.


src/extension/index.ts
```diff
@@ -0,0 +1,193 @@
...
+    const baser = new Baser(mapDistinct(fileAndUris), store);
+
+    // Panel
+    const panel = new Panel(context, baser, store);
+    disposables.push(commands.registerCommand('sarif.showPanel', () => panel.show()));
+
+    // Diagnostics
+    const diagsAll = languages.createDiagnosticCollection('SARIF');
+    const setDiags = (doc: TextDocument) => {
+        if (doc.fileName.endsWith('.git')) return;
```
sherryyshi, line 39: >doc.fileName.endsWith('.git') [](start = 12, length = 29)

can potentially do the same with files in the .git folder and the .vscode folder

src/extension/index.ts
```diff
@@ -0,0 +1,193 @@
...
+        if (store.logs.some(log => log._uri === doc.uri.toString())) return;
+        store.logs.push(...await loadLogs([doc.uri]));
+        panel.show();
+    };
+    workspace.textDocuments.forEach(syncActiveLog);
+    workspace.onDidOpenTextDocument(syncActiveLog);
+    workspace.onDidCloseTextDocument(doc => {
+        if (!doc.fileName.match(/\.sarif$/i)) return;
+        store.logs.removeWhere(log => log._uri === doc.uri.toString());
+    });
```
sherryyshi, line 72: nit: could just combine with the call backs above?

src/extension/index.ts
```diff
@@ -0,0 +1,193 @@
...
+            if (!diagnostic) return;
+
+            const result = diagnostic?.result;
+            panel.select(result);
+            if (!result) return; // Don't clear the decorations until the next result is selected.
+
+            const editor = window.visibleTextEditors.find(editor => editor.document === doc);
+            if (!editor) return; // When would editor be undef?
+
+            const locations = result?.codeFlows?.[0]?.threadFlows?.[0]?.locations ?? [];
```
sherryyshi, line 97: >? [](start = 36, length = 1)

super nit: no need due to null-check above

jeffersonking: Good eyes! Fixed.

---
In reply to: [441204378](https://github.com/microsoft/sarif-vscode-extension/pull/282#discussion_r441204378) [](ancestors = 441204378)

src/extension/index.ts
jeffersonking: One is for push(), the other is for removeWhere(). How are you proposing to merge?

---
In reply to: [441203639](https://github.com/microsoft/sarif-vscode-extension/pull/282#discussion_r441203639) [](ancestors = 441203639)

src/extension/index.ts
jeffersonking: Could you elaborate? I was only seeing spurious onDidOpen/onDidClose events for the '*.git' files so I needed to filter those out. I was not seeing any issues related to the '.git' folder and '.vscode' folder.

---
In reply to: [441201549](https://github.com/microsoft/sarif-vscode-extension/pull/282#discussion_r441201549) [](ancestors = 441201549)

src/extension/loadLogs.ts
```diff
@@ -0,0 +1,97 @@
...
+                const {data: log, pointers} = jsonMap.parse(file) as { data: Log, pointers: JsonMap};
+                log._uri = uri.toString();
+                log._jsonMap = pointers;
+                return log;
+            } catch (error) {
+                window.showErrorMessage(`Failed to parse '${uri.fsPath}'`);
+                return undefined;
+            }
+        })
+        .filter(log => log) as Log[];
```
sherryyshi, line 32: >filter(log => log) [](start = 9, length = 18)

necessary?

jeffersonking: We want to filter out the undefined array values. These value were produced on line 29. This expression is idiomatic for removing "empty" values and is used consistently across this codebase.

---
In reply to: [441209455](https://github.com/microsoft/sarif-vscode-extension/pull/282#discussion_r441209455) [](ancestors = 441209455)

src/extension/loadLogs.ts
```diff
@@ -0,0 +1,97 @@
...
+    const upgrades = logsToUpgrade.length;
+    if (upgrades) {
+        await window.withProgress(
+            { location: ProgressLocation.Notification },
+            async progress => {
+                for (const [i, oldLog] of logsToUpgrade.entries()) {
+                    if (token?.isCancellationRequested) break;
+                    progress.report({
+                        message: `Upgrading ${i + 1} of ${upgrades} log${upgrades === 1 ? '' : 's'}...`,
+                        increment: 1 / upgrades * 100
```
sherryyshi, line 45: >1 [](start = 35, length = 1)

i?

src/extension/loadLogs.ts
```diff
@@ -0,0 +1,97 @@
...
+                    });
+                    await new Promise(r => setTimeout(r, 0)); // Await needed for progress to update
+                    const {fsPath} = Uri.parse(oldLog._uri);
```
sherryyshi, line 48: would putting the `progress.report` in a promise also solve the issue?

sherryyshi: trying to understand why this is needed

---
In reply to: [441212822](https://github.com/microsoft/sarif-vscode-extension/pull/282#discussion_r441212822) [](ancestors = 441212822)

src/extension/loadLogs.ts
```diff
@@ -0,0 +1,97 @@
...
+                    try {
+                        const tempPath = upgradeLog(fsPath);
+                        const file = fs.readFileSync(tempPath, 'utf8'); // Assume scheme file.
+                        const {data: log, pointers} = jsonMap.parse(file) as { data: Log, pointers: JsonMap};
+                        log._uri = oldLog._uri;
+                        log._uriUpgraded = Uri.file(tempPath).toString();
+                        log._jsonMap = pointers;
+                        logsNoUpgrade.push(log);
+                    } catch (error) {
```
sherryyshi, line 57: >error [](start = 29, length = 5)

show the error, or send it via telemetry?

src/extension/loadLogs.ts
```diff
@@ -0,0 +1,97 @@
...
+                        window.showErrorMessage(`Failed to upgrade '${fsPath}'`);
+                    }
+                }
+            }
+        );
+    }
+    logsNoUpgrade.forEach(augmentLog);
+    if (warnUpgradeExtension) {
+        window.showWarningMessage('Some log versions are newer than this extension.');
```
sherryyshi, line 66: >'Some log versions are newer than this extension.' [](start = 34, length = 50)

Add something actionable? "Upgrade the extension to view them with the best experience" or something along those lines

src/extension/loadLogs.ts
jeffersonking: Their API wants the increment/delta as supposed to the total. I found it confusing too.

---
In reply to: [441211593](https://github.com/microsoft/sarif-vscode-extension/pull/282#discussion_r441211593) [](ancestors = 441211593)

src/extension/loadLogs.ts
jeffersonking: The user *does* need to know. As for telemetry, that feature hasn't been added yet, but we can decide then.

---
In reply to: [441213216](https://github.com/microsoft/sarif-vscode-extension/pull/282#discussion_r441213216) [](ancestors = 441213216)

src/extension/loadLogs.ts
jeffersonking: I think extensions automatically managed/updated/upgraded by VS Code. If they see this message it's likely we are behind on releases. The action would be on us.

---
In reply to: [441213743](https://github.com/microsoft/sarif-vscode-extension/pull/282#discussion_r441213743) [](ancestors = 441213743)

src/extension/loadLogs.ts
jeffersonking: Assuming it did work, would you consider that a better solution? It would still require one promise, but just in a different place, and with additional nesting.

Technically progress.report() does not require an await. You could also make the case that the rest of the function body should be wrapped in a promise. I think the the result is the same.

---
In reply to: [441212822](https://github.com/microsoft/sarif-vscode-extension/pull/282#discussion_r441212822) [](ancestors = 441212822)

jeffersonking: "trying to understand why this is needed"... really a question for VS Code. I'm guessing the the `await` allows a "thread switch" which allows the renderer thread to paint.

---
In reply to: [441213031](https://github.com/microsoft/sarif-vscode-extension/pull/282#discussion_r441213031) [](ancestors = 441213031,441212822)

jeffersonking: Comments added to clarify.

---
In reply to: [441221515](https://github.com/microsoft/sarif-vscode-extension/pull/282#discussion_r441221515) [](ancestors = 441221515,441212822)

index.html
```diff
@@ -0,0 +1,30 @@
...
+        <div id="root" style="margin: 50px 300px; position: relative;"></div><!-- relative for Popover -->
+        <script>
+            acquireVsCodeApi = (() => ({ postMessage: message => {
+                const {command, ...rest} = message;
+                if (command === 'setState') {
+                    const {state} = message;
+                    localStorage.setItem('state', state);
+                    return;
+                }
+                console.log(command, JSON.stringify(rest));
```
GabeDeBacker, line 23: >console.log [](start = 16, length = 11)

Using console.log can be super annoying to anyone else debugging their extension while this one is installed. The reason is that all console.log calls are displayed while VSCode is being debugged for all extensions (ugh). Our extension has this problem as well and I have a bug to fix it.

A technique that another team did was to add this to "launch.json"
            "name": "Launch Extension",
            ...
            "env": {
                "DEBUGGING_DEVCANVAS": "true"
            }

And they had a small wrapper around console.log which checked to see if they were in debug mode AND this env variable was set:

const gDebug: boolean = (process.execArgv.filter((v) => v.indexOf('--inspect-brk') !== -1).length > 0)
                             && (process.env.DEBUGGING_DEVCANVAS === 'true');

So, the result of this is that they only see their console.log debug spew while debugging THEIR extension. (which is nice)

.vscode/tasks.json
```diff
@@ -1,49 +1,28 @@
-// See https://go.microsoft.com/fwlink/?LinkId=733558
-// for the documentation about the tasks.json format
 {
     "version": "2.0.0",
     "tasks": [
```
GabeDeBacker, line 3: Nit: The build pipeline for this extension looks for a task to run "tslint" (which should now be eslint) which makes sure that all the code passes eslint.

Can you add back a task to run "eslint"?

I've disabled the task in our build pipeline for now, but I'd like to re-enable it.

src/extension/baser.ts
```diff
@@ -0,0 +1,171 @@
...
+        private readonly distinctLocalNames: Map<string, string>,
+        private readonly store: Pick<Store, 'distinctArtifactNames'>) {
+    }
+
+    private basesArtifactToLocal = new Map<string, string>([]) // <artifactPath, localPath>
+    private updateBases(artifact: string[], local: string[]) {
+        const i = Array.commonLength(artifact.slice().reverse(), local.slice().reverse());
+        this.basesArtifactToLocal.set(
+            artifact.slice(0, -i).join('/'),
+            local.slice(0, -i).join('/'));
```
GabeDeBacker, line 19: The '/' should be based on nodejs Path.sep.

src/extension/baser.ts
```diff
@@ -0,0 +1,171 @@
...
+import '../shared/extension';
+import { Store } from './store';
+
+export class Baser {
+    constructor(
+        private readonly distinctLocalNames: Map<string, string>,
+        private readonly store: Pick<Store, 'distinctArtifactNames'>) {
+    }
+
+    private basesArtifactToLocal = new Map<string, string>([]) // <artifactPath, localPath>
```
GabeDeBacker, line 14: >[] [](start = 59, length = 2)

Curious as to why an empty array is being passed?

GabeDeBacker, line 14: Still missing some semicolons :)

src/extension/baser.ts
```diff
@@ -0,0 +1,171 @@
...
+            }
+        }
+        return this.validatedPathsLocalToArtifact.get(localPath) ?? localPath;
+    }
+
+    // Hacky: We are using `openTextDocument` to test the existence of documents as VS Code does not provide a dedicated existence API.
+    // The similar Node `fs` API does not resolve custom URI schemes in the same way that VS Code does otherwise we would use that.
+    private async pathExists(path: string) {
+        try {
+            await workspace.openTextDocument(Uri.parse(path));
```
GabeDeBacker, line 60: using the 'strict' parameter is strongly recommended by VSCode API for all Uri.parse calls. They had to leave it optional for API backwards compatibility.

src/extension/baser.ts
```diff
@@ -0,0 +1,171 @@
...
+                this.updateValidatedPaths(artifactPath, localPath);
+                this.updateBases(artifactPath.split('/'), localPath.split('/'));
+            }
+        }
+        return this.validatedPathsLocalToArtifact.get(localPath) ?? localPath;
+    }
+
+    // Hacky: We are using `openTextDocument` to test the existence of documents as VS Code does not provide a dedicated existence API.
+    // The similar Node `fs` API does not resolve custom URI schemes in the same way that VS Code does otherwise we would use that.
+    private async pathExists(path: string) {
```
GabeDeBacker, line 58: >pathExists [](start = 18, length = 10)

Suggestions, after calling URI.parse(path, /* strict */ true), if the scheme is "file", then use fs.exist(Sync) rather than opening a document to save VSCode resources.

GabeDeBacker, line 58: >pathExists [](start = 18, length = 10)

This could be moved to a utilities file as it will probably eventually be used/needed outside of this class.s

src/extension/baser.ts
```diff
@@ -0,0 +1,171 @@
...
+            // Open Docs
+            for (const doc of workspace.textDocuments) {
+                const localPath = doc.uri.path;
+                if (localPath.file !== artifactPath.file) continue;
+                this.updateValidatedPaths(artifactPath, localPath);
+                this.updateBases(artifactPath.split('/'), localPath.split('/'));
+                return localPath;
+            }
+
+            return ''; // Can't find uri.
```
GabeDeBacker, line 116: >'' [](start = 19, length = 2)

Just curious as to why '' vs undefined?

Seems like undefined would force callers to handle the case where the string cannot be found without accidentally forgetting to check for empty string.

src/extension/baser.ts
```diff
@@ -0,0 +1,171 @@
...
+        };
+
+        let validatedUri = await validateUri();
+        if (!validatedUri && !this.activeInfoMessages.has(artifactPath)) {
+            this.activeInfoMessages.add(artifactPath);
+            const choice = await window.showInformationMessage(`Unable to find '${artifactPath.split('/').pop()}'`, 'Locate...');
+            this.activeInfoMessages.delete(artifactPath);
+            if (choice === 'Locate...') {
```
GabeDeBacker, line 124: >choice === 'Locate...' [](start = 16, length = 22)

Let's either compare the return value of showInformationMessage to undefined "if (choice)" (since there is only one argument) or defined a message item object locally can compare the return to that.

You should never compare results to a user displayable string.

src/extension/index.ts
```diff
@@ -0,0 +1,205 @@
...
+            return baser.uriBases.map(path => Uri.file(path)) as ReadonlyArray<Uri>;
+        },
+        set uriBases(values) {
+            baser.uriBases = values.map(uri => uri.path);
+        },
+    };
+}
+
+function activateDiagnostics(store: Store, baser: Baser) {
+    const diagsAll = languages.createDiagnosticCollection('SARIF');
```
GabeDeBacker, line 80: >diagsAll [](start = 10, length = 8)

You should return this value from the function and make sure it is added to the disposables array as it is disposable.

src/extension/index.ts
```diff
@@ -0,0 +1,205 @@
...
+                regionToSelection(doc, result._region),
+                result._message ?? '—',
+                severities[result.level ?? ''] ?? DiagnosticSeverity.Information, // note, none, undefined.
+                result,
+            ));
+        diagsAll.set(doc.uri, diags);
+    };
+    workspace.textDocuments.forEach(setDiags);
+    workspace.onDidOpenTextDocument(setDiags);
+    workspace.onDidCloseTextDocument(doc => diagsAll.delete(doc.uri)); // Spurious *.git deletes don't hurt.
```
GabeDeBacker, line 100: Both of these calls return disposable objects.

GabeDeBacker:
You should look through the code looking for calls into VSCode onXyz functions, they all return disposable objects which needs to be handled.

src/extension/index.ts
```diff
@@ -0,0 +1,205 @@
...
+        if (!removed.length) return;
+        window.visibleTextEditors.forEach(editor => {
+            editor.setDecorations(decorationTypeCallout, []);
+            editor.setDecorations(decorationTypeHighlight, []);
+        });
+    });
+}
+
+function activateVirtualDocuments(store: Store) {
+    workspace.registerTextDocumentContentProvider('sarif', {
```
GabeDeBacker, line 181: >registerTextDocumentContentProvider [](start = 14, length = 35)

Returns a disposable object as well

src/extension/panel.ts
```diff
@@ -0,0 +1,162 @@
...
+import { commands, ExtensionContext, TextEditorRevealType, Uri, ViewColumn, WebviewPanel, window, workspace } from 'vscode';
+import { filtersColumn, filtersRow, ResultId, _Region } from '../shared';
+import { Baser } from './baser';
+import { loadLogs } from './loadLogs';
+import { regionToSelection } from './regionToSelection';
+import { Store } from './store';
+
+export class Panel {
+    private title = 'SARIF Result'
+    @observable private panel = null as WebviewPanel | null
```
GabeDeBacker, line 15: >private panel = null as WebviewPanel | null [](start = 16, length = 43)

This would be far easier to read as
private panel: WebviewPanel | null = null;

Also, typically you should only use undefined in typescript.

src/extension/panel.ts
```diff
@@ -0,0 +1,162 @@
...
+        const doc = await workspace.openTextDocument(Uri.parse(localUri));
+        const editor = await window.showTextDocument(doc, ViewColumn.One, true);
+
+        if (region === undefined) return;
+        editor.selection = regionToSelection(doc, region);
+        editor.revealRange(editor.selection, TextEditorRevealType.InCenterIfOutsideViewport);
+    }
+
+    public select(result: Result) {
+        if (!result?._id) return; // Reduce Panel selection flicker.
```
GabeDeBacker, line 147: >!result?._id [](start = 12, length = 12)

so if _id is 0, this returns. That okay?

src/extension/panel.ts
```diff
@@ -0,0 +1,162 @@
...
+        for (const editor of window.visibleTextEditors.slice()) {
+            if (editor.document.uri.toString() !== logUri) continue;
+            await window.showTextDocument(editor.document, editor.viewColumn);
+            await commands.executeCommand('workbench.action.keepEditor');
+        }
+
+        const doc = await workspace.openTextDocument(Uri.parse(localUri));
+        const editor = await window.showTextDocument(doc, ViewColumn.One, true);
+
+        if (region === undefined) return;
```
GabeDeBacker, line 141: >region === undefined [](start = 12, length = 20)

Some places you're doing (!o) and others you are doing (o === undefined), would be nice to be consistent please.

src/shared/index.ts
```diff
@@ -0,0 +1,220 @@
...
+    }
+    for (const [key, value] of distinct) {
+        if (!value) distinct.delete(key);
+    }
+    return distinct as Map<string, string>;
+}
+
+export function augmentLog(log: Log) {
+    if (log._augmented) return;
+    log._augmented = true;
```
GabeDeBacker, line 77: This is actually more to the point I made in the original PR. I still believe it would be nice to have the "augmentations" under an object property of LOG:

    interface Log {
        augmentation? {
          _uri: string
          _uriUpgraded?: string // Only present if upgraded.
          _jsonMap?: JsonMap // Only used by the "extension" side for navigating original SARIF sources. The "panel" side does not need this feature and thus does not use this field.
          _augmented: boolean
          _distinct: Map<string, string> // Technically per Run, practically doesn't matter right now.
        }
    }

and this becomes if (log.augmentation) return

and the other code readers do not have to understand the difference between an "_" being present or not.

I'm curious about what Sherry's thoughts are here.

src/shared/extension.ts
```diff
@@ -0,0 +1,79 @@
...
+});
+
+!Array.prototype.hasOwnProperty('last') &&
+Object.defineProperty(Array.prototype, 'last', {
+    get: function() {
+        return this[this.length - 1];
+    }
+});
+
+!Array.prototype.hasOwnProperty('removeWhere') &&
```
GabeDeBacker, line 47: >removeWhere [](start = 33, length = 11)

This name is interesting because it only removes the "first" item and not "all" items that match the predicate. Can we rename this?

src/extension/panel.ts
```diff
@@ -0,0 +1,162 @@
...
+                        React.createElement(Index, { store: new Store(${JSON.stringify(state)}) }),
+                        document.getElementById('root'),
+                    )
+                </script>
+            </body>
+            </html>`;
+
+        webview.onDidReceiveMessage(async message => {
+            if (!message) return;
+            switch (message.command) {
```
GabeDeBacker, line 87: >command [](start = 28, length = 7)

Let's make sure the "command" here is casted to a strong type so in the future we don't accidentally mis-type the command name.

src/panel/details.tsx
```diff
@@ -0,0 +1,110 @@
...
+    private selectedTab = observable.box('Info')
+    @computed private get threadFlowLocations() {
+        return this.props.result?.codeFlows?.[0]?.threadFlows?.[0].locations
+            .filter(tfLocation => tfLocation.location);
+    }
+    constructor(props: DetailsProps) {
+        super(props);
+        autorun(() => {
+            const hasThreadFlows = !!this.threadFlowLocations?.length;
+            this.selectedTab.set(hasThreadFlows ? 'Code Flows' : 'Info');
```
GabeDeBacker, line 28: >'Code Flows' : 'Info' [](start = 50, length = 21)

Should strongly type the tabs.

src/panel/Table.tsx
```diff
@@ -0,0 +1,103 @@
...
+                        {store.sortColumn === col.name && <Icon title="Sort" name={store.sortDir} />}
+                        <ResizeHandle size={col.width} horizontal />
+                    </div>)}
+                </div>
+                <div tabIndex={0} className={css('svTableBody', selection.get() && 'svSelected')} onKeyDown={this.onKeyDown}>
+                    {rows.map(row => {
+                        const isSelected = selection.get() === row;
+                        if (row instanceof RowGroup) {
+                            return <Hi key={row.key} className={css('svTableRow', 'svTableCell', isSelected && 'svItemSelected')}
+                                style={{ gridTemplateColumns: '6px auto auto auto auto 1fr' /* Temp */ }}
```
GabeDeBacker, line 67: >'6px auto auto auto auto 1fr' [](start = 62, length = 29)

Should these be a style somewhere that is referenced?

src/panel/Table.tsx
```diff
@@ -0,0 +1,103 @@
...
+    columns: Column<T>[]
+    renderIconName?: (item: T) => string
+    renderGroup: (group: G) => ReactNode
+    renderCell: (column: Column<T>, itemData: T) => ReactNode
+    store: TableStore<T, G>
+}
+@observer export class Table<T, G> extends PureComponent<TableProps<T, G>> {
+    @computed get gridTemplateColumns() {
+        const {columns} = this.props;
+        return `34px ${columns.map((col, i) => `${(i === 0 ? 22 : 0) + col.width.get()}px`).join(' ')} 1fr`;
```
GabeDeBacker, line 22: >34px [](start = 16, length = 4)

Can put these values in a single place so we can easily modify the look?

src/extension/baser.ts
jeffersonking: It's extra. Now removed.

---
In reply to: [441617576](https://github.com/microsoft/sarif-vscode-extension/pull/282#discussion_r441617576) [](ancestors = 441617576)

jeffersonking: Looks like I need to add "@typescript-eslint/member-delimiter-style" on top of the rule "semi". Will do that.

---
In reply to: [441617882](https://github.com/microsoft/sarif-vscode-extension/pull/282#discussion_r441617882) [](ancestors = 441617882)

.vscode/tasks.json
jeffersonking: Can the pipeline directly do "npm run lint" (which exists)? Or must it run via a VS Code task?

---
In reply to: [441522219](https://github.com/microsoft/sarif-vscode-extension/pull/282#discussion_r441522219) [](ancestors = 441522219)

src/extension/baser.ts
jeffersonking: Agreed. Fixed.

---
In reply to: [441638778](https://github.com/microsoft/sarif-vscode-extension/pull/282#discussion_r441638778) [](ancestors = 441638778)

src/extension/panel.ts
jeffersonking: _id is a ResultId which is `[string, number, number]`, so we should be safe here.

---
In reply to: [441650171](https://github.com/microsoft/sarif-vscode-extension/pull/282#discussion_r441650171) [](ancestors = 441650171)

src/panel/details.tsx
jeffersonking: Added `type TabName = 'Info' | 'Code Flows'` and applied it.

---
In reply to: [441688737](https://github.com/microsoft/sarif-vscode-extension/pull/282#discussion_r441688737) [](ancestors = 441688737)

src/extension/panel.ts
jeffersonking: Added. Also symmetrically added it for indexStore.ts. Note, unfortunately the VS Code API to send the message isn't generic, and I haven't gone through the trouble to wrap it.

---
In reply to: [441686829](https://github.com/microsoft/sarif-vscode-extension/pull/282#discussion_r441686829) [](ancestors = 441686829)

src/shared/extension.ts
jeffersonking: Renamed to `removeFirstWhere`.

---
In reply to: [441659711](https://github.com/microsoft/sarif-vscode-extension/pull/282#discussion_r441659711) [](ancestors = 441659711)

src/panel/Table.tsx
jeffersonking: Modifying the larger look in a single place might have a few interpretations. We can discuss.

For the meanwhile, I've broken down and explained this string to make it less magical.

---
In reply to: [441694379](https://github.com/microsoft/sarif-vscode-extension/pull/282#discussion_r441694379) [](ancestors = 441694379)

src/panel/Table.tsx
jeffersonking: Removed this temporary approach.

---
In reply to: [441691371](https://github.com/microsoft/sarif-vscode-extension/pull/282#discussion_r441691371) [](ancestors = 441691371)

src/extension/panel.ts
jeffersonking: My pattern is to always use the short form, unless I can't. In this case I can't because region can be a number and 0 is a valid value.

Anecdotally the short form works 90%+ of the time and I like how it optimizes for the common case. That said, let's talk about this in the next meeting.

---
In reply to: [441653931](https://github.com/microsoft/sarif-vscode-extension/pull/282#discussion_r441653931) [](ancestors = 441653931)

src/extension/index.ts
sherryyshi: I was thinking the callbacks to `workspace.textDocuments.forEach` and `workspace.onDidOpenTextDocument` and `workspace.onDidCloseTextDocument` could be combined, but this is fine too 

---
In reply to: [441205991](https://github.com/microsoft/sarif-vscode-extension/pull/282#discussion_r441205991) [](ancestors = 441205991,441203639)

sherryyshi: As in, combined with the ones from `activateDiagnostics` 

---
In reply to: [442377153](https://github.com/microsoft/sarif-vscode-extension/pull/282#discussion_r442377153) [](ancestors = 442377153,441205991,441203639)

src/extension/index.ts
sherryyshi: ahh okay. I was just thinking there's probably no need to take any action on files opened from the .vscode and .git folder in case those were opened, but that probably won't happen that often. I didn't know there were spurious "*.git" files being "opened".

---
In reply to: [441207206](https://github.com/microsoft/sarif-vscode-extension/pull/282#discussion_r441207206) [](ancestors = 441207206,441201549)

src/extension/loadLogs.ts
sherryyshi: IMO wrapping `progress.report` in a promise (assuming that works) seems more intentional - we are letting that action complete without being blocked by the rest of the work done in this function. Just my opinion though - either way works :)

---
In reply to: [441224736](https://github.com/microsoft/sarif-vscode-extension/pull/282#discussion_r441224736) [](ancestors = 441224736,441221515,441212822)

src/extension/baser.ts
```diff
@@ -0,0 +1,171 @@
...
+            if ((	// If no workspace, then the open docs (at this moment) become the workspace.
+            // Over-assuming the localPath.name is distinct. There could be 2+ open docs with the same name.
+                !workspace.workspaceFolders?.length ||
+                    this.distinctLocalNames.has(file)
+            ) &&
+                this.store.distinctArtifactNames.has(file)) {
+
+                const artifactPath = this.store.distinctArtifactNames.get(file)!; // Not undefined due to surrounding if.
+                this.updateValidatedPaths(artifactPath, localPath);
+                this.updateBases(artifactPath.split('/'), localPath.split('/'));
```
sherryyshi, line 50: >/ [](start = 75, length = 1)

is this a file path or URI? Would this work on Windows?

jeffersonking: Theses are URI paths (soon they will be whole URIs) and are platform agnostic.

---
In reply to: [442392608](https://github.com/microsoft/sarif-vscode-extension/pull/282#discussion_r442392608) [](ancestors = 442392608)

src/extension/baser.ts
jeffersonking: Calling `openTextDocument` bugs me too and "feels" slow and this idea crossed my mind. However, in practice, I have not seen a performance concern. Are you experiencing any issues here?

Note: 1) This is effectively called on user action and is not ambiently in the background, 2) we cache so it reduces the calls to this, 3) if the file *does not* exist, I would imagine it is not meaningfully greater cost than `fs.exist`, 4) if the file *does* exist, `openTextDocument` will be called on it subsequently anyway (and VS Code seems to only open once).

---
In reply to: [441632961](https://github.com/microsoft/sarif-vscode-extension/pull/282#discussion_r441632961) [](ancestors = 441632961)

jeffersonking: Certainly possible. By default, I stick with the minimum currently (and concretely) required scope.

After replying to the above comment, I actually think it's better this method is only called when the caller knows the implications.

---
In reply to: [441633619](https://github.com/microsoft/sarif-vscode-extension/pull/282#discussion_r441633619) [](ancestors = 441633619)

src/extension/baser.ts
jeffersonking: Agreed. There are still some loose-ends where we allow non-strict URIs in from the SARIF file. I will be trying to lock those down, and then I can turn this on.

---
In reply to: [441632140](https://github.com/microsoft/sarif-vscode-extension/pull/282#discussion_r441632140) [](ancestors = 441632140)

src/extension/baser.ts
```diff
@@ -0,0 +1,171 @@
...
+        } catch (error) {
+            return false;
+        }
+        return true;
+    }
+
+    private activeInfoMessages = new Set<string>() // Prevent repeat message animations when arrowing through many results with the same uri.
+    public async translateArtifactToLocal(artifactPath: string) { // Retval is validated.
+        // Temp.
+        if (artifactPath.startsWith('sarif:')) return artifactPath;
```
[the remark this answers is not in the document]
jeffersonking: I was having some cases where the URI from the SARIF was failing to parse. Once we lock down the URIs incoming from there, I think we can switch over to uri.parse().

---
In reply to: [441637360](https://github.com/microsoft/sarif-vscode-extension/pull/282#discussion_r441637360) [](ancestors = 441637360)

index.html
jeffersonking: 100%. One of my top pet peeves. And I have been the victim of this in almost every team I've been on. :-)

That said, this file is just a tool to help debug things locally. Perhaps the comment on the top isn't obvious enough. This file is outside the `src` folder and thus never goes into the VSIX.

That's a great piece of info about the env flag. Will keep that up my sleeves for the future. :-)

---
In reply to: [441520102](https://github.com/microsoft/sarif-vscode-extension/pull/282#discussion_r441520102) [](ancestors = 441520102)

src/extension/baser.ts
jeffersonking: Both approaches could work. Ideally I should lock the return type to just `string` (and not `undefined`). This way later `if(!validatedUri)` checks below become explicit no-empty-string checks.

---
In reply to: [441634671](https://github.com/microsoft/sarif-vscode-extension/pull/282#discussion_r441634671) [](ancestors = 441634671)

src/extension/panel.ts
jeffersonking: When you say " typically you should only use undefined in typescript" is that your opinion are you referencing something?

I struggle with this. Understand the subtle semantics between the two, but I don't think I've ever seen a case where we couldn't always use just undefined. However, all the native API's (and so many external libraries) use `null` to denote absence when the pointer is a reference type. I feel the ideal is lost cause, so I aim for the next best thing: consistency with the predominant pattern.

---
In reply to: [441649046](https://github.com/microsoft/sarif-vscode-extension/pull/282#discussion_r441649046) [](ancestors = 441649046)

jeffersonking:
In the process of doing a pass.

---
In reply to: [645452749](https://github.com/microsoft/sarif-vscode-extension/pull/282#issuecomment-645452749) [](ancestors = 645452749)

src/extension/index.ts
jeffersonking: Add a comment: "When the user opens a doc, VS Code commonly silently opens the associate `*.git`. We are not interested in these events."

Not:, while it's uncommon to open files from '.vscode' and '.git', it's a legitimate scenario.

---
In reply to: [442380218](https://github.com/microsoft/sarif-vscode-extension/pull/282#discussion_r442380218) [](ancestors = 442380218,441207206,441201549)

src/extension/index.ts
jeffersonking: Ah, if we merged it, then by the time the handler is called, we wouldn't know if the `doc` was being closed or opened.

---
In reply to: [442377664](https://github.com/microsoft/sarif-vscode-extension/pull/282#discussion_r442377664) [](ancestors = 442377664,442377153,441205991,441203639)

jeffersonking: I think `syncActiveLog` might be misleading and I have renamed it to `addLog`.

---
In reply to: [442555269](https://github.com/microsoft/sarif-vscode-extension/pull/282#discussion_r442555269) [](ancestors = 442555269,442377664,442377153,441205991,441203639)

src/extension/baser.ts
jeffersonking: Found a `workspace.fs.stat` which works slightly better.

---
In reply to: [442416380](https://github.com/microsoft/sarif-vscode-extension/pull/282#discussion_r442416380) [](ancestors = 442416380,441632961)

jeffersonking: Ended up needing to move `pathExists` (now `uriExists`) to a sep for to make mocking easier. Now that we use `workspace.fs.stat` perf is not (theoretically) a concern.

---
In reply to: [442420895](https://github.com/microsoft/sarif-vscode-extension/pull/282#discussion_r442420895) [](ancestors = 442420895,441633619)

.vscode/tasks.json
jeffersonking: Resolved with Gabe offline. `npm run list` was sufficient.

---
In reply to: [441709918](https://github.com/microsoft/sarif-vscode-extension/pull/282#discussion_r441709918) [](ancestors = 441709918,441522219)

src/extension/baser.ts
jeffersonking: Now fixed (strict).

---
In reply to: [442421501](https://github.com/microsoft/sarif-vscode-extension/pull/282#discussion_r442421501) [](ancestors = 442421501,441632140)

src/shared/index.ts
jeffersonking: Discussed over chat. These underscore prefixes will eventually be removed when we exit preview status.

---
In reply to: [441658562](https://github.com/microsoft/sarif-vscode-extension/pull/282#discussion_r441658562) [](ancestors = 441658562)

src/extension/baser.ts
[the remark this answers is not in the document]
jeffersonking: Fixed.

---
In reply to: [442422296](https://github.com/microsoft/sarif-vscode-extension/pull/282#discussion_r442422296) [](ancestors = 442422296,441637360)

src/extension/baser.ts
jeffersonking: Constrained the typing to just `string`.

---
In reply to: [442427277](https://github.com/microsoft/sarif-vscode-extension/pull/282#discussion_r442427277) [](ancestors = 442427277,441634671)

src/extension/baser.ts
jeffersonking: Fixed.

---
In reply to: [441707908](https://github.com/microsoft/sarif-vscode-extension/pull/282#discussion_r441707908) [](ancestors = 441707908,441617882)

src/extension/panel.ts
jeffersonking: Changed this to colon notation.

---
In reply to: [442430959](https://github.com/microsoft/sarif-vscode-extension/pull/282#discussion_r442430959) [](ancestors = 442430959,441649046)

jeffersonking:
Complete.

---
In reply to: [646259322](https://github.com/microsoft/sarif-vscode-extension/pull/282#issuecomment-646259322) [](ancestors = 646259322,645452749)

PR #283, opened 2020-06-30
Migrate Stacks code from previous codebase

mukul-ms:
Goal:
- We added code to display Stacks property as a Tab in previous codebase. In this PR, we are trying to migrate it here 

Scenarios:
- Single + Multiple Stacks
image

- No Stack
image

### DIFF
--- a/src/panel/details.tsx
+++ b/src/panel/details.tsx
@@ -45,7 +45,7 @@ interface DetailsProps { result: Result, height: IObservableValue<number> }
 
         const {result, height} = this.props;
         const helpUri = result?._rule?.helpUri;
-        const renderItem = (location: Location) => {
+        const renderLocation = (location: Location) => {
 			const { message, uri, region } = parseLocation(result, location)
 			return <>
 				<div className="ellipsis">{message ?? '—'}</div>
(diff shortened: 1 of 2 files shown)


src/panel/details.tsx
```diff
@@ -40,6 +45,25 @@ interface DetailsProps { result: Result, height: IObservableValue<number> }
 
         const {result, height} = this.props;
         const helpUri = result?._rule?.helpUri;
+        const renderItem = (location: Location) => {
```
GabeDeBacker, line 48: >renderItem [](start = 14, length = 10)

can we rename this to renderLocation for consistency with renderStack?

src/panel/details.tsx
```diff
@@ -81,28 +105,49 @@ interface DetailsProps { result: Result, height: IObservableValue<number> }
...
-                                    <div className="ellipsis">{message ?? '—'}</div>
-                                    <div className="svSecondary">{uri?.file ?? '—'}</div>
-                                    <div className="svLineNum">{region?.startLine ?? '—'}:1</div>
-                                </>;
-                            };
+                                const location = change.newValue
+                                postSelectArtifact(result, location?.physicalLocation)
+                            })
 
                             return <List items={items} renderItem={renderItem} selection={selection} allowClear>
```
GabeDeBacker, line 118: >renderItem [](start = 67, length = 10)

let's change this to renderLocation please

src/panel/details.tsx
mukul-ms: Fixed

src/panel/details.tsx
mukul-ms: Fixed

PR #284, opened 2020-07-02
Add tests for tableStore

mukul-ms:


package.json
```diff
@@ -88,7 +88,7 @@
         "prestart": "npm install",
         "start": "webpack --watch",
         "server": "webpack-dev-server -d",
-        "test": "mocha",
+        "test": "nyc --reporter=text mocha",
```
jeffersonking, line 91: We should maintain the ability to run the "fast test loop".

jeffersonking: Also, where is `nyc` added as a package? If it was installed globally, then we should make it local.

src/panel/tableStore.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+        const tableStore2 = new TableStore(groupBy, itemSource2, selection);
+        const oddRowGroup2 = tableStore2.rows[0] as RowGroup<number, string>;
+        assert.deepEqual(oddRowGroup2.items.map((item) => item.item), [5,3,1]);
+
+        const itemSource3 = { results: [21,4,19,2,13,1,5] }
+        const tableStore3 = new TableStore(groupBy, itemSource3, selection);
+        const oddRowGroup3 = tableStore3.rows[0] as RowGroup<number, string>;
+        assert.deepEqual(oddRowGroup3.items.map((item) => item.item), [21,19,13,1,5]);
+    });
+    it('Effect of toggling sort', () => {
```
jeffersonking, line 80: What is being tested here? `'test_column_name'` is not a real column.

src/panel/tableStore.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+import { _Region } from '../shared';
+import assert from 'assert';
+
+describe('TableStore', () => {
+    it('Get rows based on different values of expanded property', () => {
+        // Create instance of tableStore
+        const groupBy = (item: number) => item % 2 == 0 ? 'even' : 'odd';
+        const itemSource = { results: [1,2,3,4,5] }
+        const selection = observable.box(); 
+        const tableStore = new TableStore(groupBy, itemSource, selection);
```
jeffersonking, line 15: We should find a way to not copy/paste lines 12-15 in this file.

src/panel/tableStore.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+        assert.strictEqual((tableStore.rows[0] as RowGroup<number,string>).title, 'odd');
+
+        // tableStore - when # of even elements are more than # of odd elements
+        const itemSource2 = { results: [1,2,3,4,5,6,8,10] }
+        const tableStore2 = new TableStore(groupBy, itemSource2, selection);
+
+        // "even" row group would be sorted high in the list
+        assert.strictEqual((tableStore2.rows[0] as RowGroup<number,string>).title, 'even');
+    })
+    it ('Row groups maintain the same sequence of row items as it is in itemSource', () => {
```
jeffersonking, line 62: `RowGroup.items`, as I mentioned before, is considered opaque.

package.json
```diff
@@ -153,5 +153,8 @@
     "optionalDependencies": {
         "@microsoft/sarif-multitool-darwin": "2.2.5",
         "@microsoft/sarif-multitool-win32": "2.2.5"
-    }
+    },
+    "nyc": {
+		"exclude": "src/**/*.spec.ts src/test/*" 
+	}
```
jeffersonking, line 159: Spaces instead of tabs.

mukul-ms: Fixed

package.json
mukul-ms: 1) Added `nyc` to dev-dependency in package.json 
2) for `to maintain fast test loop` since `--exit` option is not provided and `watch` is set to `true` inside mocha config, so to see the nyc report we would need to exit the mocha process. Sorry what's the recommendation?

src/panel/tableStore.spec.ts
mukul-ms: I wanted to test the `toggleSort` and how it effects `rows` but its a fair point sort doesn't do much until implemented by a child class. So I'm removing this test case for now. Will test sorting logic when I write tests for `resultTableStore`

package.json
jeffersonking: It's okay to have a separate script for each.

mukul-ms: Oh that's a good point. Thanks will do that

src/panel/tableStore.spec.ts
mukul-ms: Moved `groupBy` and `selection` to outside the `it` blocks but inside `describe` block. I can probably create a utility method which creates a tableStore instance based on `results` provided (since thats the only thing changing between tests). Any suggestion?

package.json
mukul-ms: Fixed

src/panel/tableStore.spec.ts
mukul-ms: So I'm assuming not something to test against?

mukul-ms: Cool. Fixed

Issue #293, opened 2020-07-08
File URIs should be displayed "nicely" (decoded) to the user in the viewer..


GabeDeBacker:
While working with on of our interns, it was clear to me that our display of file URIs in the UI is a bit confusing to the user (and when copied, cannot be opened - at least in Windows).

The issue is that the file URIs are escaped (for example the colon in c:/  is escaped to c%3a/ etc.

The suggest is to display file URIs, well, not as URIs but there correct (non-escaped) format.

for example:
Instead of displaying "file://c%3a/users/gabrield/repos/foo.cpp" if the URI scheme is a file, show "c:\users\gabrield\repos\foo.cpp" instead. If the URI is not a file, leave it alone.

This will make more sense to most of our users and will allow the URI to be copy\pasted into things like notepad, file explorer, etc. (i.e. integrate better with the developer workflows)

Christellah:
We noticed this issue in two places: in the Logs tab of the viewer and in the tooltip of the info tab when an error is selected, as shown on the screenshot below.

![image](https://user-images.githubusercontent.com/25352470/87190531-a373b500-c2c0-11ea-8cbd-fbd0f9e6cf8c.png)

jeffersonking:
I think this (in technical terms) means: If the URI scheme is "file", then render `fsPath`. Otherwise maintain current behavior: render the full string (`toString`) of the URI.

mukul-ms:
- Current status: Was able to reproduce, working on the fix. Issue found at 2 locations in the UI


-  ![image](https://user-images.githubusercontent.com/43248455/87351125-f77fd300-c526-11ea-82e6-1e74455d7726.png)

- ![image](https://user-images.githubusercontent.com/43248455/87351198-154d3800-c527-11ea-87ad-7d6df60d47d7.png)

PR #301, opened 2020-07-13
Show decoded file URIs

mukul-ms:
Fix for https://github.com/microsoft/sarif-vscode-extension/issues/293 

1) For `Log` field inside `Info` Panel
- Before: 
  - image

- After 
  - image

2) For `Log` tab
- Before: 
   - image

- After
   - image

### DIFF
--- a/src/extension/loadLogs.ts
+++ b/src/extension/loadLogs.ts
@@ -21,7 +21,7 @@ export async function loadLogs(uris: Uri[], token?: { isCancellationRequested: b
                 const file = fs.readFileSync(uri.fsPath, 'utf8')  // Assume scheme file.
                     .replace(/^\uFEFF/, ''); // Trim BOM.
                 const {data: log, pointers} = jsonMap.parse(file) as { data: Log, pointers: JsonMap};
-                log._uri = decodeURIComponent(uri.toString());
+                log._uri = uri.scheme === 'file' ? uri.fsPath : uri.toString();
                 log._jsonMap = pointers;
                 return log;
             } catch (error) {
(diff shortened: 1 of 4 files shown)


jeffersonking:
There is still a URI scheme after the fix. You probably want to take a look at my comment on the issue.

mukul-ms:
> There is still a URI scheme after the fix. You probably want to take a look at my comment on the issue.

Thanks, addressed the feedback.

src/extension/loadLogs.ts
```diff
@@ -21,7 +21,7 @@ export async function loadLogs(uris: Uri[], token?: { isCancellationRequested: b
                 const file = fs.readFileSync(uri.fsPath, 'utf8')  // Assume scheme file.
                     .replace(/^\uFEFF/, ''); // Trim BOM.
                 const {data: log, pointers} = jsonMap.parse(file) as { data: Log, pointers: JsonMap};
-                log._uri = uri.toString();
+                log._uri = uri.scheme === 'file' ? uri.fsPath : uri.toString();
```
jeffersonking, line 24: There is a less invasive way to fix this issue than to change the underlying schema.

mukul-ms: Pushed a fix. Please let me know if it works

src/panel/details.tsx
```diff
@@ -63,8 +64,10 @@ interface DetailsProps { result: Result, height: IObservableValue<number> }
 				<div className="ellipsis">{text ?? '—'}</div>
 				<div className="svSecondary">{uri?.file ?? '—'}</div>
 				<div className="svLineNum">{region?.startLine}:1</div>
-			</>;
-		};
```
jeffersonking, line 67: Outside of this PR scope, but what's going on with the whitespace here?

mukul-ms: Weird. pushed a fix in this PR

mukul-ms:
> Recommend refactoring the duplicate code here at some point.

Pushed a fix in this PR

PR #315, opened 2020-08-04
Add Azure Monitor (aka Application Insights)

jeffersonking:
Just basic integration for now. No events reported aside from the "free ones": activation and errors.

The CSS fix is an unrelated rider.

### DIFF
--- a/src/extension/telemetry.ts
+++ b/src/extension/telemetry.ts
@@ -1,0 +1,16 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+import TelemetryReporter from 'vscode-extension-telemetry';
+import { publisher, name, version } from '../../package.json';
+
+let reporter: TelemetryReporter;
+
+export function activate() {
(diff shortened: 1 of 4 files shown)


src/extension/telemetry.ts
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+import TelemetryReporter from 'vscode-extension-telemetry';
+import { publisher, name, version } from '../../package.json';
+
+let reporter: TelemetryReporter;
+
+export function activate() {
+    const key = 'bf8e52c4-6749-4709-92a0-e3a8fd589648';
```
mukul-ms, line 10: should we have a seperate config to store keys like these? I'm assuming so far we only work with 1 so seperate file might be overkill?

jeffersonking: There's no intention of ever changing/configuring this, so I think it should be ok.

Note https://github.com/microsoft/vscode-cpptools and https://github.com/microsoft/sarif-website do the same.

mukul-ms: Sounds good, thanks

PR #320, opened 2020-08-06
Fix driver-less rule grouping.

jeffersonking:
See code comments for best explanation. Roy run into this bug.


jeffersonking:
Pressed for time so bypassing the `Sarif Viewer` checks.

PR #332, opened 2020-08-14
Misc bug fixes.

jeffersonking:
Addresses a perf issue where a 150MB log takes 14s to load. With this fix it takes 3.5s. Memory pressure is no longer limited at 500MB. It seems to be smooth over 1.5GB now (no heap issues).

We tradeoff the ability to navigate to the result region. With at 150MB, we probably don't need that anyway.

Rider: 3 misc fixes.

Attaching our internal mail thread here:

I investigated (profiled) Roy's 150MB SARIF log. On my (slow) machine it takes ~14s to load. 10.5s of that is generating the source map. We can (situationally) drop the source map and shave 75% of the time.

What is the source map for? When you click the "6101_1.sarif" link below, it takes to you the region in the SARIF Log source. The source map powers that. A feature primarily for SARIF producers, not consumers.

Screen Shot 2020-08-14 at 2 02 16 AM

A few options to situationally disable the source map:
* Auto-disable on large files. Not great as many small files can still cause trouble.
* Add a user setting. I don't think it will be discoverable and we would still need smart defaults.
* Lazy-create the source map when that link is clicked. User may be surprised there is a long wait. Also makes the code more-stateful, which I recommend against.
* Auto-disable on API-triggered SARIF log opens. I'm recommending this one.

Thoughts? If nobody cares (say in 24 hours), I'll go with my recommendation. Thanks.

### DIFF
--- a/.github/workflows/prerelease.yaml
+++ b/.github/workflows/prerelease.yaml
@@ -18,7 +18,6 @@ jobs:
             ${{ runner.os }}-node-
       - run: |
           npm install
-          npm install --force @microsoft/sarif-multitool-darwin @microsoft/sarif-multitool-win32
           npx json -I -f package.json -e 'this.version = `${this.version}-${{ github.run_number }}`'
           npm run package
       - id: package_version
--- a/src/extension/loadLogs.ts
+++ b/src/extension/loadLogs.ts
@@ -4,12 +4,11 @@
 /// <reference path="jsonSourceMap.d.ts" />
 import { execFileSync } from 'child_process';
 import * as fs from 'fs';
-import jsonMap from 'json-source-map';
 import { Log, ReportingDescriptor } from 'sarif';
 import { eq, gt, lt } from 'semver';
 import { tmpNameSync } from 'tmp';
 import { ProgressLocation, Uri, window } from 'vscode';
-import { augmentLog, JsonMap } from '../shared';
+import { augmentLog } from '../shared';
 
 const driverlessRules = new Map<string, ReportingDescriptor>();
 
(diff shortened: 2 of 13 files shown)


.github/workflows/prerelease.yaml
```diff
@@ -18,7 +18,6 @@ jobs:
             ${{ runner.os }}-node-
       - run: |
           npm install
-          npm install --force @microsoft/sarif-multitool-darwin @microsoft/sarif-multitool-win32
```
mukul-ms, line 21: sorry why are we removing installing multitool again?

jeffersonking:
@roywill If you think it makes more sense to lazy-generate the source map on demand. We can do that instead.

src/extension/loadLogs.ts
```diff
@@ -13,14 +13,23 @@ import { augmentLog, JsonMap } from '../shared';
...
+function jsonParse(text: string, skipPointers?: 'skipPointers'): { data: Log, pointers: JsonMap | undefined } {
+    return skipPointers
+        ? {
+            data: JSON.parse(text),
+            pointers: undefined,
+        }
+        : jsonMap.parse(text) as { data: Log, pointers: JsonMap };
+}
+
+export async function loadLogs(uris: Uri[], skipPointers?: 'skipPointers', token?: { isCancellationRequested: boolean }) {
```
mukul-ms, line 25: Sorry what's the reasoning behind having an option for `skipPointers` flag? From what it seems like right now regardless of file size we won't be generating source maps right? 

Is it related to the question we asked to Roy about ability to having it on demand?

mukul-ms: I just read the email thread, it gave more context. Thanks, resolving this comment now

.github/workflows/prerelease.yaml
jeffersonking: Rationale on #319. Hopefully that answers it.

jeffersonking:
I reverted the main gist of this PR which is about "Perf: Disable SARIF log goto result region on API-opened logs." in favor another PR.

Now just merging the leftover bug fixes.

PR #337, opened 2020-08-19
Fix #330

### LINKED ISSUE #330: [question] how to do path mapping in VSCode about
Hi Sarif team,

I have output the .sarif file, but when click the display result, can't go to the source file, how to do path mapping about this:

```

Locations: | \%SRCROOT%\common\script\ruijiecommon.py (15)
-- | --
Related: | \%SRCROOT%\common\script\ruijiecommon.py (242)


```

企业微信截图_15973972264791

企业微信截图_15973987932087

jeffersonking:
Fixes #330.

I looked at the the schema in the bug (`https://raw.githubusercontent.com/oasis-tcs/sarif-spec/master/Schemata/sarif-schema-2.1.0.json`) and it seems to be equal to `rtm5`. Thus we don't need to run it through the Multitool.

### DIFF
--- a/src/extension/loadLogs.ts
+++ b/src/extension/loadLogs.ts
@@ -82,7 +82,9 @@ export function detectUpgrade(log: Log, logsNoUpgrade: Log[], logsToUpgrade: Log
             ?.replace('http://json.schemastore.org/sarif-', '')
             ?.replace('https://schemastore.azurewebsites.net/schemas/json/sarif-', '')
             ?.replace(/\.json$/, '');
-        if (schema === undefined || schema === '2.1.0-rtm.5') {
+        if (schema === undefined || schema === '2.1.0-rtm.5'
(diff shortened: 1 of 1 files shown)


jeffersonking:
@lgolding Add you as a reviewer to double-check my assertion about this schema.

lgolding:
You are correct. Those schemas are identical for our purposes. The differences are:

- The title/id/description properties at the top.
- The schema in the SARIF spec repo declares that it conforms to JSON Schema draft-07; the one in the SDK repo declares that it conforms to draft-04. That's because the VS JSON editor doesn't (or didn't, at the time) process draft-07 schemas.

So you don't need to run it through the Multitool (I assume you mean its `transform` command, which would update pre-release versions).

---
In reply to: [675854830](https://github.com/microsoft/sarif-vscode-extension/pull/337#issuecomment-675854830) [](ancestors = 675854830)

PR #347, opened 2020-09-01
Preliminary implementation of "fixes" (colloquially QuickFix) .

jeffersonking:
"Preliminary" because this is the first time (afaik) this aspect of the SARIF spec is being implemented. There are certain scenarios (such as cascading fixes) that still require some discussion. Nevertheless, it's still good to get what we have so far into the hands of the team.

A refactoring/cleanup or the regions logic was a prerequisite.

### DIFF
--- a/src/extension/index.ts
+++ b/src/extension/index.ts
@@ -156,7 +156,7 @@ function activateDecorations(disposables: Disposable[], store: Store, panel: Pan
                 const text = tfl.location?.message?.text;
                 return `Step ${i + 1}${text ? `: ${text}` : ''}`;
             });
-            const ranges = locations.map(tfl => regionToSelection(doc, parseRegion(tfl.location?.physicalLocation?.region)));
+            const ranges = locations.map(tfl => regionToSelection(doc, tfl.location?.physicalLocation?.region));
             const rangesEnd = ranges.map(range => {
                 const endPos = doc.lineAt(range.end.line).range.end;
                 return new Range(endPos, endPos);
--- a/src/extension/index.ts
+++ b/src/extension/index.ts
@@ -47,7 +47,7 @@ export async function activate(context: ExtensionContext) {
     // Borrowed from: https://github.com/Microsoft/vscode-languageserver-node/blob/db0f0f8c06b89923f96a8a5aebc8a4b5bb3018ad/client/src/main.ts#L217
     const isDebugOrTestMode =
         process.execArgv.some(arg => /^--extensionTestsPath=?/.test(arg)) // Debug
-        && process.execArgv.some(arg => /^--(debug|debug-brk|inspect|inspect-brk)=?/.test(arg)); // Test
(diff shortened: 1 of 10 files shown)


src/extension/index.ts
```diff
@@ -109,6 +111,52 @@ function activateDiagnostics(disposables: Disposable[], store: Store, baser: Uri
     disposables.push(workspace.onDidCloseTextDocument(doc => diagsAll.delete(doc.uri))); // Spurious *.git deletes don't hurt.
     const disposer = observe(store.logs, () => workspace.textDocuments.forEach(setDiags));
     disposables.push({ dispose: disposer });
+    return setDiags;
+}
+
+function activateCodeActions(disposables: Disposable[], baser: UriRebaser, refreshDiagnostics: (document: TextDocument) => void) {
```
GabeDeBacker, line 117: Just my opinion, but having 2 functions nested inside this larger function makes this very difficult for me to read :(. I would assume that the "fix function" is going to get a lot bigger, so why not split it out into another function or even another file?

src/extension/index.ts
```diff
@@ -109,6 +111,52 @@ function activateDiagnostics(disposables: Disposable[], store: Store, baser: Uri
...
+            const localUri = await baser.translateArtifactToLocal(artifactUri);
+            for (const replacement of replacements) {
+                edit.replace(
+                    Uri.parse(localUri),
+                    regionToSelection(document, replacement.deletedRegion),
+                    replacement.insertedContent?.text ?? ''
+                );
+            }
+        }
+        workspace.applyEdit(edit);
```
GabeDeBacker, line 135: > workspace.applyEdit(edit); [](start = 7, length = 27)

How does applying this edit affect the other "replacements" in the SARIF log? Do they need to be adjusted somehow?

src/extension/index.ts
```diff
@@ -109,6 +111,52 @@ function activateDiagnostics(disposables: Disposable[], store: Store, baser: Uri
...
+        result._fixed = true;
+        refreshDiagnostics(document);
+    }));
+
+    const provider = new class implements CodeActionProvider {
```
GabeDeBacker, line 140: >CodeActionProvider [](start = 42, length = 18)

So why not move this implementation to its own file?

src/extension/index.ts
jeffersonking: Yes they would need to be adjusted. This touches on my leading PR comment "There are certain scenarios (such as cascading fixes) that still require some discussion." We should discuss this at our next meeting.

src/extension/index.ts
jeffersonking: Discussed offline. In short there's some question to the viability of this feature without significant additional investment in "tracking spans". Thus keeping these lines of code "packed-up and ready to go" just in case.

src/extension/index.ts
jeffersonking: Discussed offline. For other reviewers, see first thread above about "2 functions nested".

PR #348, opened 2020-09-02
Drop support for log upgrades. Warn instead.

jeffersonking:
Our upgrade process has always been a thorn. The Multitool binaries constitute over 99% of our VSIX weight. Upgrading cannot be done in memory (only on disk). Also, we had some platform-specific failures that were difficult to reproduce.

I discussed with @michaelcfanning, and the SARIF ecosystem advocates using the latest version (2.1.0 rtm5). The long term plan is not to perpetually have the viewers support older versions. There will always be ways outside of the viewer to upgrade these files.

Thus dropping support for upgrade in this PR. We will be listening for customer feedback and have a plan in place in case it is largely negative. On the other hand, if this works out, I think it will be good for everyone.

Rider: I fixed up some of the unit tests.

### DIFF
--- a/src/extension/loadLogs.ts
+++ b/src/extension/loadLogs.ts
@@ -2,12 +2,10 @@
 // Licensed under the MIT License.
 
 /// <reference path="jsonSourceMap.d.ts" />
-import { execFileSync } from 'child_process';
-import * as fs from 'fs';
+import { readFileSync } from 'fs';
 import { Log, ReportingDescriptor } from 'sarif';
 import { eq, gt, lt } from 'semver';
-import { tmpNameSync } from 'tmp';
-import { ProgressLocation, Uri, window } from 'vscode';
+import { Uri, window } from 'vscode';
 import { augmentLog } from '../shared';
 import * as Telemetry from './telemetry';
 
@@ -18,7 +16,7 @@ export async function loadLogs(uris: Uri[], token?: { isCancellationRequested: b
         .map(uri => {
             if (token?.isCancellationRequested) return undefined;
             try {
-                const file = fs.readFileSync(uri.fsPath, 'utf8')  // Assume scheme file.
+                const file = readFileSync(uri.fsPath, 'utf8')  // Assume scheme file.
                     .replace(/^\uFEFF/, ''); // Trim BOM.
                 const log = JSON.parse(file) as Log;
                 log._uri = uri.toString();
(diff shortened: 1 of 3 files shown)


src/extension/loadLogs.ts
```diff
@@ -33,42 +31,19 @@ export async function loadLogs(uris: Uri[], token?: { isCancellationRequested: b
...
+        const {fsPath} = Uri.parse(log._uri, true);
+        window.showWarningMessage(`'${fsPath}' was not loaded. Version '${log.version}' and schema '${log.$schema ?? ''}' is not supported.`);
     }
-    logsNoUpgrade.forEach(log => augmentLog(log, driverlessRules));
+    logsSupported.forEach(log => augmentLog(log, driverlessRules));
     if (warnUpgradeExtension) {
         window.showWarningMessage('Some log versions are newer than this extension.');
     }
-    return logsNoUpgrade;
+    return logsSupported;
```
lgolding, line 46: >logsSupported [](start = 11, length = 13)

Would you mind calling it `supportedLogs`?

src/extension/loadLogs.ts
```diff
@@ -33,42 +31,19 @@ export async function loadLogs(uris: Uri[], token?: { isCancellationRequested: b
...
 }
 
 export function detectUpgrade(log: Log, logsNoUpgrade: Log[], logsToUpgrade: Log[]) {
```
lgolding, line 49: >{ [](start = 84, length = 1)

As long as we have the hood up, I suggest we improve the messaging. Let's have `detectUpgrade` take `supportedLogs`, `oldVersionLogs`, `unrecognizedSchemaLogs`, and `futureVersionLogs`. Place each log in exactly one of those lists. Don't bother with the Boolean return value, which loses the information about which logs were too new. Don't collapse the messaging for version mismatch and schema mismatch.

Once you've processed all the logs and the arrays are filled, then loop over each of `oldVersionLogs`, `unrecognizedSchemaLogs`, and `futureVersionLogs`, and produce the appropiate message:
- `'{fsPath}' was not loaded because this extension does not support the obsolete SARIF version '{log.version}'.`
- `'{fsPath}' was not loaded because this extension does not recognize the schema '{log.$schema}'.`
- `'{fsPath}' was not loaded because its version '{log.version}' is newer than the version '2.1.0' supported by this extension. Consider upgrading to a new version of the extensions.`

lgolding (changes requested):
:clock1:

src/extension/loadLogs.ts
jeffersonking: How frequently do we expect the end user to leverage the finer information? My working experience is that most folks using this don't appreciate schema vs version. This message is essentially an opaque string meant to be forwarded back to us (the developers).

Strictly speaking the end user doesn't even have to open an issue as we have telemtry covering this also.

lgolding (approved):
:shipit:

src/extension/loadLogs.ts
jeffersonking: Likewise with logsNotSupported? I might be weird but prefixes (as opposed to suffixes) help my mind group similar things. The visual alignment "clicks" for me.

lgolding: Up to you. In my commented below I suggested `supportedLogs`, `oldVersionLogs`, `unrecognizedSchemaLogs`, and `futureVersionLogs`.  But you're the one that has to read the code every day.

---
In reply to: [483873846](https://github.com/microsoft/sarif-vscode-extension/pull/348#discussion_r483873846) [](ancestors = 483873846)

src/extension/loadLogs.ts
jeffersonking: Let's discuss offline when you're back.

GabeDeBacker (approved):
:shipit:

PR #349, opened 2020-09-03
Update icon, readme, and description.

jeffersonking:
In prep for v3 release.


src/extension/index.ts
```diff
@@ -42,8 +42,7 @@ export async function activate(context: ExtensionContext) {
     disposables.push(commands.registerCommand('sarif.showPanel', () => panel.show()));
 
     // General Activation
-    const refreshDiagnostics = activateDiagnostics(disposables, store, baser);
```
lgolding, line 45: >activateDiagnostics [](start = 31, length = 19)

Why are we no longer doing this?

README.md
```diff
@@ -1,66 +1,40 @@
...
+## What's New?
+
+Version 3 incorporates many feedback-based improvements:
+* Improved keyboard accessibility within the **SARIF Results Panel**. Arrow key through the list of results.
+* Resizable details section within the **SARIF Results Panel**.
+* Generally improved performance and responsiveness.
+* Automatic reconciliation of URIs between the SARIF log and your local workspace in most cases.
+
+To focus our efforts, we have dropped some less-used and less-reliable features:
+* Support for old SARIF versions - We now strictly support the current version (2.1.0 rtm.5) and version back (2.1.0 rtm.4). Older versions can be upgraded with the standalone SARIF Multitool (via [nuget](https://www.nuget.org/packages/Sarif.Multitool/) and [npm](https://www.npmjs.com/package/@microsoft/sarif-multitool)).
```
lgolding, line 19: >(2.1.0 rtm.5) and version back (2.1.0 rtm.4) [](start = 79, length = 44)

Don't even mention it. Just say "2.1.0". If they load an rtm.4 file and we don't complain, that's just gravy.

lgolding (approved):
:shipit:

src/extension/index.ts
jeffersonking: refreshDiagnostics was a "hook" to allow CodeActions to call. Removing the feature removes the need for this.

README.md
```diff
@@ -71,21 +45,18 @@ await sarifExt.exports.openLogs([
...
+
+## Development
 
-## Use
-1. Open a .sarif file
-2. Results will show up the Problems Panel
-3. Click the result you're investigating:
-    * The editor will navigate to the location
-    * The Sarif Explorer will open with the result details
+`F5` launches this extension. Subsequent changes are watched and rebuilt. Use command `workbench.action.reloadWindow` to see the changes.
```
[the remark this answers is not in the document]
lgolding: Wait, are you talking about debugging the extension? If so, why do we have to tell them about F5? Isn't it the same gesture when debugging any code base?

---
In reply to: [483872893](https://github.com/microsoft/sarif-vscode-extension/pull/349#discussion_r483872893) [](ancestors = 483872893)

README.md
lgolding: Oh, and it's not just the "current" version. We don't want people wondering when the "next" version is coming out. Let's say "support the public standard version 2.1.0".

---
In reply to: [483872733](https://github.com/microsoft/sarif-vscode-extension/pull/349#discussion_r483872733) [](ancestors = 483872733)

jeffersonking: We do not support 2.1.0 rtm.0 through rtm.3. I assume you're implying it's still okay to claim "support the public standard version 2.1.0". Doesn't seem people will care.

README.md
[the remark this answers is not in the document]
jeffersonking: Perhaps one of these facts may justify it?:
- Outside of the Windows and Full Visual Studio ecosystem, I'm not sure if F5 is guaranteed. I suspect F5 is not the norm for the common GitHub project.
- We did have to custom author a `launch.json` for F5 to work (with Webpack). So this wasn't "free".
- We also hooked up `npm install` to precede 'F5`. Typically you would need to manually install first.
- Perhaps "F5 launches this extension." is too obvious. But it English-structurally-conveniently sets the stage for the next two less obvious sentences.

If you have a suggested phrasing, I'm down to change this. :-)

GabeDeBacker (approved):
:shipit:

PR #353, opened 2020-09-10
Refine prerelease pipeline auto-versioning.

jeffersonking:
Previously our Github Action would use the `github.run_number` as the `prerelease` number. The flaw was that it did not reset to zero when the main version changed. We now calculate the `prerelease` number based on past tags and the package version.


GabeDeBacker (approved):
:shipit:

PR #352, opened 2020-09-08
Add a "Remove All Logs" button

sivadeilra:
When working with a large number of log files, it can be tedious to
click "x" on each log file. This adds a button that removes (closes)
all log files at once.

This resolves #351 .

PR #356, opened 2020-09-15
Add a "Close All Logs" button (revised).

### LINKED ISSUE #351: Need easy way to close all log files
When working with large numbers of SARIF logs, it's tedious to click "x" on every single log file to remove it from the view. It would be good to have a single "Close All Logs" button.

I have a PR ready which implements this; creating the issue to track it.

jeffersonking:
Revision of #352 which fixes #351.

I took what @sivadeilra started and added conditional visibility. The "close all logs" button now shares a spot with the "expand/collapse all" button. Rationale:
- We're running low on horizontal space thus rationing. We also don't quite have enough for more than one row.
- We don't need "expand/collapse all" button for logs. Conversely "close all logs" is a bit destructive and limiting it to the Logs tab conveniently mitigates accidental clicks.
- It's unfortunate that the iconography differences are super subtle in the two buttons. But we can always adjust pending feedback.

I renamed "remove" to "close" to better mirror the analogous APIs.

Rider: Disabled some tests that @mukul-ms may need to update.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
 	"name": "sarif-viewer",
-	"version": "3.0.0",
+	"version": "3.0.2",
 	"lockfileVersion": 1,
 	"requires": true,
 	"dependencies": {
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
     "description": "Adds support for viewing SARIF logs",
     "author": "Microsoft Corporation",
     "license": "MIT",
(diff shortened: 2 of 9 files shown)


sivadeilra (approved):
Looks great.  Thanks!

GabeDeBacker (approved):
:shipit:

GabeDeBacker (approved):

PR #359, opened 2020-09-23
Fix #357: Display non-failure results appropriately

### LINKED ISSUE #357: Display non-failure results appropriately.
When a result has `kind` equal to `"review"`, the extension displays it as "informational". But in fact, a `"review"` result should be displayed as "warning", because it means that it's an issue that requires human review to determine if there's actually a problem.

The guidance for viewers is:

- Errors: results that break/block processes.
- Warnings: items actively tracked that should always display to users but not block anything.
- Informational: low-importance items that do *not* display by default.

Here is how the viewer should treat results with a `kind` value _other than_ `"fail"`:

| `kind` value | Viewer treatment |
| --- | --- |
| `"informational"` | Informational |
| `"notApplicable"` | Informational |
| `"open"` | Warning [1] |
| `"pass"` | Informational [2] |
| `"review"` | Warning |

[1] Because this value means that the tool didn't have enough information to know if there was a problem. This is used by program correctness provers. If a prover doesn't know (for example), if a particular function can throw an exception, it might not be able to decide whether a code path is safe.

[2] Unless the extension allows you to show a green check-mark or the like for a "pass" result.

@michaelcfanning FYI

lgolding:
See description of #357 for the spec.

@michaelcfanning FYI

### DIFF
--- a/src/shared/index.ts
+++ b/src/shared/index.ts
@@ -151,7 +151,7 @@ export function effectiveLevel(result: Result): Result.level {
         case 'review':
             return 'warning';
         case 'fail':
-        default:    // default case should never happen.
+        default:
             return result.level ?? result._rule?.defaultConfiguration?.level ?? 'warning';
     }
 }
--- a/src/shared/index.ts
+++ b/src/shared/index.ts
@@ -126,7 +126,7 @@ export function augmentLog(log: Log, rules?: Map<string, ReportingDescriptor>) {
             result._message = format(message.text || result.message?.text, result.message.arguments) ?? '—';
             result._markdown = format(message.markdown || result.message?.markdown, result.message.arguments); // No '—', leave undefined if empty.
 
-            result.level = result.level ?? result._rule?.defaultConfiguration?.level ?? 'warning';
+            result.level = effectiveLevel(result);
             result.baselineState = result.baselineState ?? 'new';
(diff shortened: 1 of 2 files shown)


src/shared/index.ts
```diff
@@ -141,6 +141,21 @@ export function augmentLog(log: Log, rules?: Map<string, ReportingDescriptor>) {
...
+    switch (result.kind) {
+        case 'informational':
+        case 'notApplicable':
+        case 'pass':
+            return 'note';
+        case 'open':
+        case 'review':
+            return 'warning';
+        case 'fail':
+        default:    // default case should never happen.
```
jeffersonking, line 154: Confused by why "default case should never happen". I often seen `kind` undefined.

lgolding: I may be wrong about that comment.

In the .NET SDK, the deserialization is controlled by a set of property-level attributes such that if `Kind` is missing, it will populate `fail` into the object model (it's a bit more complicated, but that's the idea). So in .NET there will always be a valid value. Not so here, I take it.

---
In reply to: [493825383](https://github.com/microsoft/sarif-vscode-extension/pull/359#discussion_r493825383) [](ancestors = 493825383)

jeffersonking (approved):
Some feedback was giving over a call (and not reflected in comments).

PR #391, opened 2021-07-06
Create codeql-analysis.yml

eddynaka:


jeffersonking:
@eddynaka Merged within 5 minutes with no description and no review?

eddynaka:
SDL compliance task.
Adding to all repos that we are missing.

jeffersonking:
I think it's important note that (SDL compliance) initially in the PR (or the YAML). Not all contributors may be aware of that.

How does this overlap or not overlap with the existing LGTM integration?

eddynaka:
Both LGTM and CodeQL uses the same queries.

CodeQL is more integrated with the Security tabs and we can run based on crons.
LGTM is more integrated with PRs. So, you will see the issues once you create a PR.

IF your project gets stale, for example, and a vulnerability starts to exist, LGTM would not notify you, since it only works based on PRs.
Now, CodeQL works on crons and PRs (merged), when that happens, we would receive the notifications.

jeffersonking:
> LGTM is more integrated with PRs. So, you will see the issues once you create a PR.

I'm seeing the equivalent functionality with the CodeQL action. Can you quantify the "more integrated" part? If the CodeQL action is a superset, then do we need both?

eddynaka:
> > LGTM is more integrated with PRs. So, you will see the issues once you create a PR.
> 
> I'm seeing the equivalent functionality with the CodeQL action. Can you quantify the "more integrated" part? If the CodeQL action is a superset, then do we need both?

If we have an issue during PR, for example, an unused variable. During PR, LGTM would say: 'you have an unused variable, you can remove'. Since CodeQL can take time to execute, it's not advisable to run during PR.

Below, all the rules that it ran:
- https://github.com/microsoft/sarif-vscode-extension/pull/391/checks?check_run_id=2998930639
- https://lgtm.com/projects/g/microsoft/sarif-vscode-extension/logs/languages/lang:javascript

"more integrated": once it finds an issue, it will post results here: https://github.com/microsoft/sarif-vscode-extension/security/code-scanning.

About the rules, per CodeQL repository, they share the same set of rules:
```
This open source repository contains the standard CodeQL libraries and queries that power LGTM and the other CodeQL products that GitHub makes available to its customers worldwide.
```

references:
- https://github.com/github/codeql

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,71 @@
...
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ main, dev ]
+  pull_request:
```
jeffersonking, line 17: > Since CodeQL can take time to execute, it's not advisable to run during PR.

jeffersonking: So do we need both?

eddynaka: I see.
The template changed from my last time.
The previous was:
```
on:
  push:
    branches: [ main ]
  schedule:
    - cron: '30 23 * * 0'
```

So, in the case that I was imagining, it wasn't running during PRs.
Based on that we can:
either remove LGTM or remove pull_request analysis from CodeQL.

jeffersonking: I'll defer to your recommendation.

PR #392, opened 2021-08-06
Use ReportingDescriptorReference for rule lookup

cannist:
When looking up a rule for a result so far the viewer only used `result.ruleIndex` or attempted to find the rule in the driver by its id. This PR adds additional lookup logic that supports results having a ReportingDescriptorReference ([see spec §3.52](https://docs.oasis-open.org/sarif/sarif/v2.1.0/csprd01/sarif-v2.1.0-csprd01.html#_Toc10541300)). This is especially important if rules are listed under an extension which would not have been found with the previous implementation.

### Motivation
GitHub CodeQL, a producer of SARIF files, added the option to list rules under `runs[].tool.extensions` instead of `run[].tool.driver` in [version 2.5.1](https://github.com/github/codeql-cli-binaries/releases/tag/v2.5.1) of its command line tools to give more accurate information about the origin of a rule implementation. SARIF files created by [CodeQL Action](https://github.com/github/codeql-action) use this format by default.
Currently the viewer shows no rule information for results in SARIF files that use this format.

### Restrictions of this implementation:
The spec also specifies lookups via a `guid` field which I have not tried to cover.


cannist:
@GabeDeBacker This has been sitting here for while. Is there anything I can do to move it along?

jeffersonking:
@cannist I will review this tomorrow. Thanks for your patience!

src/shared/index.ts
```diff
@@ -118,9 +118,23 @@ export function augmentLog(log: Log, rules?: Map<string, ReportingDescriptor>) {
...
+            // Case 4: result has no ReportingDescriptorReference but a ruleIndex
+            //         Component defaults to driver and we look up the rule via ruleIndex
+            // Case 5: We make a last effort to look up the rule via id in the driver
+            // Case 6: Give up and create a rule without metadata
+            result._rule = run.tool.extensions?.[result?.rule?.toolComponent?.index ?? -1]?.rules?.[result?.rule?.index ?? -1] // 1
+                ?? run.tool.extensions?.[result?.rule?.toolComponent?.index ?? -1]?.rules?.[result?.ruleIndex ?? -1] // 2
+                ?? run.tool.driver.rules?.[result?.rule?.index ?? -1] // 3
+                ?? run.tool.driver.rules?.[result.ruleIndex ?? -1] // 4
+                ?? run.tool.driver.rules?.find(rule => rule.id === result.ruleId) // 5
+                ?? getDriverlessRule(result.ruleId); // 6
```
jeffersonking, line 137: I recommend refactoring for clarity:

```ts
const toolComponent
    =  run.tool.extensions?.[result?.rule?.toolComponent?.index ?? -1]
    ?? run.tool.driver;

result._rule
    =  toolComponent.rules?.[result?.rule?.index ?? result.ruleIndex ?? -1]
    ?? toolComponent.rules?.find(rule => rule.id === result.ruleId)
    ?? getDriverlessRule(result.ruleId);
```

This has the side effect of supporting an id-based lookup in an extension. I don't have the SARIF spec memorized, but I believe that is desirable behavior.

jeffersonking (changes requested):

jeffersonking:
@cannist If convenient, can you share (and check-in) a sample SARIF file that represents your scenario here? I'd like to capture that for use for future testing. Thanks!

samples/rulesExtensions.sarif
```diff
@@ -0,0 +1,188 @@
...
+                  "startLine": 3
+                }
+              }
+            }
+          ],
+          "message": {
+            "text": "Unused variable foo."
+          }
+        },
+        {
```
cannist, line 138: Lookup tool component via `guid` and rule via `index`. This does not work at the moment.

samples/rulesExtensions.sarif
```diff
@@ -0,0 +1,188 @@
...
+                  "startLine": 4
+                }
+              }
+            }
+          ],
+          "message": {
+            "text": "Unused variable foo."
+          }
+        },
+        {
```
cannist, line 161: Lookup tool component via `index` and rule via `index`. This works.

cannist (commented):
I've refactored according to your suggestions, @jeffersonking .

I've also added a version of one of our test files that covers several rule lookup cases. It has more cases than the code I've added supports... let me know if you'd rather have these excluded. I'm explaining each case inline.

jeffersonking (approved):

jeffersonking:
@cannist Approved and merged. Thanks for responding to that feedback and attaching the sample file. Also thanks for your patience!

Issue #397, opened 2021-09-02
Issue line numbers in panel off by one 


huornlmj:
Clicking on the issues shown to be on specific lines numbers (e.g. issue shown as being on line 230), highlights the code but one line earlier (line 229)

![image](https://user-images.githubusercontent.com/28444739/131819925-fa3a916e-adeb-4c31-bcb0-ea227c90d34a.png)

Issue #373, opened 2021-02-26
Line numbers are off by one (version 3.0.2 still)


viennadd:
Seems it is an opposite situation with related issue #300 

with use of [off-by-one.sarif.txt](https://github.com/microsoft/sarif-vscode-extension/files/4913792/off-by-one.sarif.txt) in #300 , the SARIF result located at line 16, but it shows at line 17.
![image](https://user-images.githubusercontent.com/3396288/109317264-80b20300-7887-11eb-8bc0-8e6be3140d9e.png)

viennadd:
I think we don't need to convert from 0-based to 1-based at here, because SARIF line numbers are 1-based already.
https://github.com/microsoft/sarif-vscode-extension/blob/bd13e8021549fb1674795a9bd473803318dfe673/src/shared/index.ts#L118-L120

viennadd:
seems we don't have to convert from 0-based to 1-based at here, because SARIF line numbers are 1-based already.
https://github.com/microsoft/sarif-vscode-extension/blob/bd13e8021549fb1674795a9bd473803318dfe673/src/shared/index.ts#L118-L120

MichaelSquires:
I've been seeing this issue as well.

eddynaka:
Hello,

sorry for the delay.
We fixed this in version 3.1.0.

The PR:
https://github.com/microsoft/sarif-vscode-extension/pull/399

PR #399, opened 2021-09-29
Fix line number offset by 1

lin-ll:
This PR addresses these open issues:
- https://github.com/microsoft/sarif-vscode-extension/issues/397
- https://github.com/microsoft/sarif-vscode-extension/issues/373

The existing code assumes that the line number passed from a sarif result location is zero-indexed. However, the line numbers are already 1-indexed. See https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc34317690 which describes `startLine`, for example, as a positive integer. 

See also https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc34317409, which indicates that line numbers are 1-based.

### DIFF
--- a/src/shared/index.spec.ts
+++ b/src/shared/index.spec.ts
@@ -33,7 +33,7 @@ describe('augmentLog', () => {
     it('add augmented fields', () => {
         augmentLog(log);
         assert.strictEqual(result._uri, 'file:///folder/file.txt');
-        assert.strictEqual(result._line, -1);
+        assert.strictEqual(result._line, 0);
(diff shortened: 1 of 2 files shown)


jeffersonking (approved):
Thanks for contributing @lin-ll!

The `_line` actually is no longer needed. Historically it was more relevant. These days we can just short-circuit directly to `_region`. I will submit a follow-up PR, but approving this anyway.

jeffersonking:
@lin-ll Nope. Thanks!

Issue #400, opened 2021-09-30
Result Properties in the Panel


lin-ll:
At one point, the sarif viewer showed result properties in the panel. However, it looks as though this functionality was commented out when the new UI was introduced in this [PR](https://github.com/microsoft/sarif-vscode-extension/pull/282). Not sure if that was intentional?

PR #401, opened 2021-09-30
Display properties in the details of the panel

lin-ll:
This uncomments the line so that result properties can be shown in the details of the panel. 

See https://github.com/microsoft/sarif-vscode-extension/issues/400

### DIFF
--- a/src/panel/details.tsx
+++ b/src/panel/details.tsx
@@ -100,7 +100,11 @@ interface DetailsProps { result: Result, height: IObservableValue<number> }
                                                                 }}>
                                                                 {result._log._uri.file}{result._log._uriUpgraded && ' (upgraded)'}
                                                             </a>
-                            {/* <span>Properties</span>		<span><pre><code>{JSON.stringify(selected.properties, null, '  ')}</code></pre></span> */}
+                            <span>Properties</span>		    <span>
+                                                                <pre>
+                                                                    <code>{JSON.stringify(result.properties, null, 2)}</code>
+                                                                </pre>
+                                                            </span>
                         </div>
                     </div>
                 </Tab>


jeffersonking:
What are the specific properties you are trying to display?

Just checking because often exists a SARIF construct better suited for the job. And when there isn't, it is useful feedback for the next version of SARIF. Properties are intended for exceptional cases.

lin-ll:
> What are the specific properties you are trying to display?
> 
> Just checking because there often exists a SARIF construct better suited for the job. And when there isn't, it is useful feedback for the next version of SARIF. Properties are intended for exceptional cases.

I see! I happen to be working with some SARIF results that have a lot of metadata, like associated names, statistics, etc. that don't seem suitable for the other SARIF Result fields. For me, it would be really useful to be able to see the PropertyBag in the panel so I don't have to dig through them.

jeffersonking:
@lin-ll Thanks for you patience! As it stands in this PR, the Property Bag items would look like this:

Screen Shot 2021-11-01 at 5 13 30 PM

I would like to propose integrating these items more seamlessly like this:

Screen Shot 2021-11-01 at 5 14 11 PM

I have the code for this in-hand. Thoughts? Would this fit your needs? Thanks!

lin-ll:
> @lin-ll Thanks for you patience! As it stands in this PR, the Property Bag items would look like this:
> 
> Screen Shot 2021-11-01 at 5 13 30 PM
> 
> I would like to propose integrating these items more seamlessly like this:
> 
> Screen Shot 2021-11-01 at 5 14 11 PM
> 
> I have the code for this in-hand. Thoughts? Would this fit your needs? Thanks!

@jeffersonking - that looks excellent and would definitely fit my needs. Thank you!

jeffersonking:
@lin-ll: Created #407 with the discussed changes. Will close this one in favor of that.

PR #407, opened 2021-11-02
Render "Property Bags" in result details.

jeffersonking:
Alternative to PR #401 (Issue #400). Instead of rendering plain JSON, this approach renders the Property Bag like this:

Screen Shot 2021-11-01 at 5 14 11 PM

### DIFF
--- a/src/panel/details.tsx
+++ b/src/panel/details.tsx
@@ -6,7 +6,7 @@
 import { autorun, computed, IObservableValue, observable } from 'mobx';
 import { observer } from 'mobx-react';
 import * as React from 'react';
-import { Component } from 'react';
+import { Component, Fragment } from 'react';
 import ReactMarkdown from 'react-markdown';
 import { Location, Result, StackFrame } from 'sarif';
 import { parseArtifactLocation, parseLocation, decodeFileUri } from '../shared';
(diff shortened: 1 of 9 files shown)


eddynaka:
@yongyan, we should do something similar for VS

src/panel/details.tsx
```diff
@@ -100,7 +100,29 @@ interface DetailsProps { result: Result, height: IObservableValue<number> }
...
+                                    {Object.entries(rest).map(([key, value]) => {
+                                        return <Fragment key={key}>
+                                            <span className="ellipsis">{key}</span>
+                                            <span>{(() => {
+                                                if (value === null)
+                                                    return '—';
+                                                if (Array.isArray(value))
+                                                    return <span style={{ whiteSpace: 'pre' }}>{value.join('\n')}</span>;
+                                                if (typeof value === 'boolean')
+                                                    return JSON.stringify(value, null, 2);
```
eddynaka, line 117: do we need to use `null, 2`? In this case, isn't this similar to JSON.stringify(value)? The same comment applies for the next stringify.

eddynaka: or...wait...looking at the docs, it will add two spaces at the end. If that's correct, is that really needed?

jeffersonking: The intent was to have pretty-printed JSON and the `null, 2` was for that.

eddynaka (approved):

src/panel/details.tsx
jeffersonking: "2" means "indent with 2 spaces". It is not adding spaces at the end.

sherryyshi (approved):

PR #434, opened 2022-04-22
Vulnerable package updates

EasyRhinoMSFT:


jeffersonking:
Not sure we should be adding explicit dependencies (`json-schema` and `markdown-it`) as a method of pinning the version.

Per a previous mail (see March 7), some of these can be dismissed as "Risk is tolerable". I have reviewed the risks.

PR #435, opened 2022-05-05
Add option for GHAS integration

yongyan-gh:
# Description

The option allows users to enable/disable the GHAS feature.

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,6 +42,11 @@
                     "type": "boolean",
                     "default": true
                 },
+                "sarif-viewer.enableGhasIntegration": {
+                    "description": "Indicates whether to enable GitHub Advanced Security integration.",
+                    "type": "boolean",
+                    "default": false
+                },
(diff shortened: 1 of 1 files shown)


jeffersonking:
The `main` branch does not support GHAS. Currently only the `ghas` branch.

yongyan-gh:
> The `main` branch does not support GHAS. Currently only the `ghas` branch.

so I should add this in ghas branch?

jeffersonking:
If you have installing private builds from the `ghas` branch you implicitly want this feature on. Thus we don't need the switch for that branch.

Before `ghas` gets merged to main, we need to decide what mechanism enables/disables the feature. Or perhaps it's always on. As far as I know, the design has not been decided yet (the requirements have not all arrived yet).

jeffersonking:
Closing this due to staleness. We will revisit this when the design of the feature is settled.

Issue #427, opened 2022-03-08
CodeQL CLI Sarif Mismatch


theemanofsteele:
I've discovered that the Sarif file that the CodeQL CLI produces is a schema URL that causes the sarif-vscode-extension to think the file is invalid. The schema URl (valid) that the CodeQL CLI produces looks like this: 
`"$schema" : "https://json.schemastore.org/sarif-2.1.0.json"`
Which would produce the following error:
`'\results.sarif' was not loaded. Version '2.1.0' and schema 'https://json.schemastore.org/sarif-2.1.0.json' is not supported.`

So I debugged the extension and discovered a couple items: 
Please refer to this block of code: https://github.com/microsoft/sarif-vscode-extension/blob/main/src/extension/loadLogs.ts#L55-L67
- Line 57 is actually `http` and not `https` . In this specific case, `http://json.schemastore.org/sarif-2.1.0.json` and `https://json.schemastore.org/sarif-2.1.0.json` both still resolve. But CodeQL uses `https` where as the sarif-vscode-extension is using `http`. 
- `https://raw.githubusercontent.com/oasis-tcs/sarif-spec/master/Schemata/sarif-schema-2.1.0` No longer resolves. Possibly a dead path?
- `https://docs.oasis-open.org/sarif/sarif/v2.1.0/cos02/schemas/sarif-schema-2.1.0` is still valid and so I compared it to `https://json.schemastore.org/sarif-2.1.0.json` and they are identical with one exception: oasis-open has a typo in a description. 
![image](https://user-images.githubusercontent.com/17327021/157171416-cd2eefa5-ecb3-49a1-bbd7-42c0e88e3726.png)
So possibly `https://json.schemastore.org/sarif-2.1.0.json` is safe to add to the current list?

I am happy to open a PR with some changes if ya'll want. But thought I'd check in first before doing anything. 😃 
Thanks!

theemanofsteele:
Looks like all the Oasis URLs broke. Mentioned here: https://github.com/oasis-tcs/sarif-spec/issues/509
but 👇 might be the fix for the current broken Oasis URL
http://docs.oasis-open.org/sarif/sarif/v2.1.0/os/schemas/sarif-schema-2.1.0.json

vovikhangcdv:
Same issue.

theemanofsteele:
Looks like both of the urls on lines 61 and 62 were broken since first submitting this issue. The proposed change addresses that.

[...]

Issue #422, opened 2022-02-23
This extension does not load the logs from qodana


fredericgboutin-yapla:
Context: https://www.jetbrains.com/help/qodana/qodana-sarif-output.html

So yeah, I'm not using PhpStorm since I prefer VSCode but their Php linter is interesting. Long story short, I'm using quodana to generatea a SARIF file which I currently cannot open with this extension without modifying the `$schema` property.

I have to change the initial value from `https://raw.githubusercontent.com/schemastore/schemastore/master/src/schemas/json/sarif-2.1.0-rtm.5.json` to `https://schemastore.azurewebsites.net/schemas/json/sarif-2.1.0-rtm.5.json`

Cheers!

vovikhangcdv:
Seem like the same issue is here https://github.com/microsoft/sarif-vscode-extension/issues/427

PR #436, opened 2022-05-06
Relax `$schema` Contraints

jeffersonking:
Addresses #422 and #427. Treating a wider range of `$schema` values as valid.

### DIFF
--- a/src/extension/loadLogs.spec.ts
+++ b/src/extension/loadLogs.spec.ts
@@ -51,30 +51,30 @@ describe('loadLogs', () => {
         assert.strictEqual(logs.every(log => log.version === '2.1.0'), true);
     });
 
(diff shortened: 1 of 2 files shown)


EasyRhinoMSFT:
We might want to hold on this -- I talked to Michael and he wants to put it in front of the TC for discussion. He's offline for some reason but I'll ping him as soon as he's available so we can get this moving.

jeffersonking:
@EasyRhinoMSFT @michaelcfanning Any word from the TC? @nixel2007 and some others are blocked by this bug.

FWIW: I don't think this PR paints us into a corner. At least not anymore than we already are.

EasyRhinoMSFT (approved):
:shipit:

Issue #443, opened 2022-06-15
Some documents will not have decorations when inspecting multi-file code flows


keremc:
This appears to be a side effect of #442. I think [this line](https://github.com/microsoft/sarif-vscode-extension/blob/7b7bfc4d9d423b52ac217d0f6319228691ae2b73/src/extension/index.ts#L170) throws when a thread-flow location has a range that cannot be valid in the active document. As a result, you will not see any decorations in such a document.

Issue #442, opened 2022-06-15
Code-flow decorations are displayed in the wrong file


keremc:
When inspecting a result with a code flow, SARIF Viewer displays decorations for all the thread flows, even if they are supposed to be in a different file. You can download [this sample here](https://github.com/microsoft/sarif-vscode-extension/files/8910058/sample.tar.gz) and check it out yourself. When you select `Branch condition evaluates to a garbage value` in the result list, it renders a decoration for `Entered call from 'selcall'` in `amd64/sysv.c`.

https://user-images.githubusercontent.com/5321759/173848616-a6dd8f9a-e187-41ba-ac2d-ae75eeb6be78.mp4

PR #444, opened 2022-06-15
Improve decorations

### LINKED ISSUE #441: The step decorations jump between code flows in some cases
This seems to happen when the currently selected thread-flow object and another unrelated result with a code flow point to the same line in a file. The decoration messages unexpectedly switch to the other result when you click the thread-flow location, as you can see in the attached video.

The culprit here appears to be [the code actions provider](https://github.com/microsoft/sarif-vscode-extension/blob/7b7bfc4d9d423b52ac217d0f6319228691ae2b73/src/extension/index.ts#L156). It is triggered when I jump to line 667 and resets the decorations.

I'm attaching a sample, so you can reproduce the issue: [sample.tar.gz](https://github.com/microsoft/sarif-vscode-extension/files/8909880/sample.tar.gz).

https://user-images.githubusercontent.com/5321759/173840265-b8518c70-dad1-4b73-bae1-2cd53485c9a4.mp4

keremc:
This PR should fix #441, #442 and #443. I added `selection` observable to the `Panel` class to keep track of the result in the main extension component. The decorations are now set by an observer instead of the code actions provider. Unfortunately, the logic for finding the relevant document for an artifact location is a bit fragile. If you have any suggestions, I would like to hear them.

### DIFF
--- a/src/extension/index.ts
+++ b/src/extension/index.ts
@@ -54,7 +54,7 @@ export async function activate(context: ExtensionContext) {
     // General Activation
     activateDiagnostics(disposables, store, baser);
     activateWatchDocuments(disposables, store, panel);
-    activateDecorations(disposables, store);
+    activateDecorations(disposables, store, panel);
     activateVirtualDocuments(disposables, store);
     activateSelectionSync(disposables, panel);
 
@@ -140,7 +140,7 @@ function activateWatchDocuments(disposables: Disposable[], store: Store, panel: 
 }
 
 // Decorations are for Call Trees.
-function activateDecorations(disposables: Disposable[], store: Store) {
+function activateDecorations(disposables: Disposable[], store: Store, panel: Panel) {
     const decorationTypeCallout = window.createTextEditorDecorationType({
         after: { color: new ThemeColor('problemsWarningIcon.foreground') }
     });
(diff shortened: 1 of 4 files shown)


jeffersonking:
I have coincidentally been working fixing the same issues you mentioned in #442 and #443. I have not had a chance to review your changes, but I wanted to give a heads up first to prevent potentially duplicate work.

jeffersonking:
Aside from the conflict with #445, my main concern with this PR is that the `select` command is not a "`selectResult`" command. The `select` command serves use cases beyond selecting Results. If you look at the callers to `postSelectArtifact` (which fires the `select` command) you will see these other cases. Assuming intent to select a Result will have unintended side effects. For example, when someone clicks on an embedded link (see `renderMessageTextWithEmbeddedLinks`), they will now have an unwanted Result selection which may result in an unwanted decoration activation.

My other concern is that values such as `logUri` have been removed without consideration of downstream impact. Because of this `selectLocal(logUri...` will receive undefined and Logs will not longer be pinned as before.

keremc:
> Aside from the conflict with https://github.com/microsoft/sarif-vscode-extension/pull/445, my main concern with this PR is that the select command is not a "selectResult" command. The select command serves use cases beyond selecting Results. If you look at the callers to postSelectArtifact (which fires the select command) you will see these other cases. Assuming intent to select a Result will have unintended side effects.

Thanks for the feedback. I thought adding a result ID field to the `select` message would be fine because, at least
as far as I can tell, it is only ever generated by `postSelectArtifact` which always takes a `Result` as the first
argument. But I guess (in addition to the issue with `renderMessageTextWithEmbeddedLinks`) this can change in
the future and you would rather not limit the use cases of `select`.

I'm wondering if it would be OK to include the result ID as an optional field.

> For example, when someone clicks on an embedded link (see renderMessageTextWithEmbeddedLinks), they will now have an unwanted Result selection which may result in an unwanted decoration activation.

Additionally, with my PR applied, the decorations no longer update after selecting a diagnostic in the "Problems" panel. If I'm not mistaken, VS Code doesn't offer a way of detecting such interactions with the problem panel, so we have to use `registerCodeActionsProvider`.

PR #445, opened 2022-06-21
Analysis Steps improvements and other improvements.

jeffersonking:
Note: I recommend reviewing by commit rather than the cumulatively.

The largest change is with Analysis Steps (Thread Flows). Previously flows were constrained to a single document. Now if you select a multi-document flow, and happen to have several documents open, you will see the decorations simultaneously light up. Indentation was also added to the Analysis Steps list to help visualize stack depth. Especially useful in long flows.

The next largest change better leveraging base URIs. This should offer an improved experience (less "Locate file..." dialogs) for one of our most common cases: where the Log contains base URIs and the user has a corresponding workspace open.

A few more changes are incoming, but I wanted to get a head start on the review process.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "sarif-viewer",
-    "version": "3.1.2",
+    "version": "3.2.0",
     "lockfileVersion": 2,
     "requires": true,
     "packages": {
         "": {
             "name": "sarif-viewer",
-            "version": "3.1.2",
+            "version": "3.2.0",
             "license": "MIT",
             "dependencies": {
                 "follow-redirects": "1.14.8",
--- a/src/extension/index.activateDecorations.ts
+++ b/src/extension/index.activateDecorations.ts
@@ -1,0 +1,79 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+/* eslint-disable filenames/match-regex */
+
+import { IArraySplice, observe } from 'mobx';
+import { Log } from 'sarif';
+import { Disposable, languages, Range, ThemeColor, window } from 'vscode';
+import { parseArtifactLocation } from '../shared';
+import '../shared/extension';
+import { regionToSelection } from './regionToSelection';
+import { ResultDiagnostic } from './resultDiagnostic';
+import { Store } from './store';
+
+// Decorations are for Analysis Steps.
+export function activateDecorations(disposables: Disposable[], store: Store) {
+    const decorationTypeCallout = window.createTextEditorDecorationType({
+        after: { color: new ThemeColor('problemsWarningIcon.foreground') }
+    });
+    const decorationTypeHighlight = window.createTextEditorDecorationType({
+        border: '1px',
+        borderStyle: 'solid',
+        borderColor: new ThemeColor('problemsWarningIcon.foreground'),
+    });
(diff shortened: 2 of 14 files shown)


package-lock.json
```diff
@@ -1,12 +1,12 @@
...
     "name": "sarif-viewer",
-    "version": "3.1.2",
+    "version": "3.2.0",
     "lockfileVersion": 2,
     "requires": true,
     "packages": {
         "": {
             "name": "sarif-viewer",
-            "version": "3.1.2",
+            "version": "3.2.0",
```
marmegh, line 9: >version": "3.2.0", [](http://example.com/codeflow?start=13&length=18)

Coming in new here, I'm reading this correctly that this is just version tracking for the extension. Correct?

src/extension/index.ts
```diff
@@ -98,7 +103,27 @@ function activateDiagnostics(disposables: Disposable[], store: Store, baser: Uri
         // When the user opens a doc, VS Code commonly silently opens the associate `*.git`. We are not interested in these events.
         if (doc.fileName.endsWith('.git')) return;
 
-        const artifactUri = baser.translateLocalToArtifact(doc.uri.toString(true /* skipEncoding */));
+        const artifactUri = (() => {
+            // TODO: Recall why skipEncoding=true in the common case.
+            // TODO: Review for consistent usage of skipEncoding (for example in `provideTextDocumentContent`)
```
marmegh, line 108: Can we get issues to track these and note the issues in the comment?

package-lock.json
jeffersonking: Yes, correct.

src/extension/index.ts
jeffersonking: Done: #448

src/extension/index.activateDecorations.ts
```diff
@@ -0,0 +1,116 @@
...
+            const result = diagnostic?.result;
+            if (!result) return; // Don't clear the decorations. See `activeResultId` comments.
+
+            activeResultId.set(JSON.stringify(result._id)); // Stringify for comparability.
+
+            // Technically should be using `onDidChangeTextEditorSelection` and `languages.getDiagnostics`
+            // then manually figuring with diagnostics are at the caret. However `languages.registerCodeActionsProvider`
+            // provides the diagnostics for free. The only odd part is that we always return [] when `provideCodeActions` is called.
+            return [];
+        }
```
keremc, line 46: The code actions provider still causes unexpected switches between results when going through the analysis steps in some cases. (see #441) Wouldn't it make more sense to make the panel webview include the result ID in the `select` messages and set `activeResultId` in the message handler? I'm not experienced with VS Code extensions, so maybe there's something I'm missing.

EasyRhinoMSFT (approved):
:shipit:

src/extension/index.activateDecorations.ts
jeffersonking: @keremc The root issue is when multiple Results overlap the same range of text we do not support selecting a specific Result within that set. I put more details in #441.

The `select` messages are used for more than just selecting Results. If we are to include additional data, I would prefer to account for all the cases.

Also, `activeResultId` is local to the Decorations feature. `activeResultId` is created out of convenience as an event source. The data it represents is derived, and normally I would prefer not to store this.

PR #454, opened 2022-08-15
Block inline scripts

jeffersonking:
Mitigating potential malicious SARIF Log embedded links. Addressing this by removing "unsafe-line" on the meta CSP. "Inline" in this context means not only links, but also script blocks. This eliminates a range of attack vectors.

The downstream effect is that one of our our own inline script blocks needs to be externalized (into `init.js`). Externalizing it requires some changes to the packaging settings. Thus this PR affects more files than you would imagine.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "sarif-viewer",
-    "version": "3.2.0",
+    "version": "3.2.2",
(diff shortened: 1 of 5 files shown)


EasyRhinoMSFT (approved):
:shipit:

PR #462, opened 2022-09-27
Less-obtrusive first-time flow.

jeffersonking:
With the last round of GitHub Code Scanning changes, the UX got too aggressive (in hindsight). For previous users of this extension with no knowledge of GitHub Code Scanning, they would suddenly get the viewer opened. Very often (statically speaking) there would be no results to show.

Attempting to be less obtrusive (while still promoting the new feature) by only showing the panel if Code Scanning results are found. This also means the progress indicator, which normally is displayed by the panel, needs an alternate place to go (we added a progress card).


jeffersonking:
Merging without review as @EasyRhinoMSFT is on vacation. Will follow up when he is back.

PR #475, opened 2022-12-05
Watch files opened via API

### LINKED ISSUE #467: Automatically reflect changes to SARIF files in the viewer
If a SARIF file is open in the viewer, any changes to that file should be automatically reflected.

Scenario:
1. Static analysis run produces a SARIF file; SARIF Viewer is opened
2. After fixing some of the issues, the static analysis tool is run again and updates/overwrites the existing SARIF file

Expected behavior: the SARIF viewer automatically reflects the changes in the file.

Actual behavior: nothing changes :(.

50Wliu:
Implements & closes #467.

This is a very simple file-watching implementation - for every log opened via `api.openLogs()`, we'll add it to a file watcher. When those files change (or are deleted), we'll update the SARIF Viewer as appropriate. We'll remove the files from the watcher once they're closed.

What I don't know is how chokidar handles multiple `add` calls for the same file - will it start watching the same file _twice_? Or is it smart enough to know that it's already watching and no-op? I don't have an easy way to test that scenario, but I'm hoping it's a pretty rare edge case for someone to repeatedly `openLogs` on a file that's already open.

I've tested this locally and confirmed that the openLogs implementation works, but not that closeLogs successfully removes from the watcher (though I'd be very surprised if it didn't).

### DIFF
--- a/src/extension/index.ts
+++ b/src/extension/index.ts
@@ -200,13 +200,13 @@ function activateVirtualDocuments(disposables: Disposable[], store: Store) {
             if (contents?.rendered?.text) return contents?.rendered?.text;
             if (contents?.text) return contents?.text;
             if (contents?.binary) {
-                const lines = Buffer.from(contents?.binary, 'base64').toString('hex').match(/.{1,32}/g) ?? [];
-                return lines.reduce((sum, line, i) => {
+                const lines = Buffer.from(contents?.binary, 'base64').toString('hex').match(/.{1,32}/g);
+                return lines?.reduce((sum, line, i) => {
                     const lineNo = ((i + 128) * 16).toString(16).toUpperCase().padStart(8, '0');
                     // eslint-disable-next-line no-control-regex
                     const preview = Buffer.from(line, 'hex').toString('utf8').replace(/(\x09|\x0A|\x0B|\x0C|\x0D|\x1B)/g, '?');
                     return `${sum}${lineNo}  ${line.toUpperCase().match(/.{1,2}/g)?.join(' ')}  ${preview}\n`;
-                }, '');
+                }, '') ?? '';
             }
             token.isCancellationRequested = true;
             return '';
--- a/src/extension/index.ts
+++ b/src/extension/index.ts
@@ -70,6 +70,13 @@ export async function activate(context: ExtensionContext) {
 
     // For now, we keep file watch functionality limited to the API.
     // After we're sure it's working well, consider also enabling it for files opened manually.
+    // FIXME: Possible race condition if a file is changed multiple times in quick succession.
+    // 1. First change event fires (denoted as [1]).
+    // 2. [1] Log is removed from the store.
+    // 3. Second change event fires (denoted as [2]).
+    // 4. [2] Log is removed from the store (returns false as it's already been removed).
+    // 5. [1] Log is re-loaded and added to the store.
+    // 6. [2] Log is re-loaded and added (again!) to the store.
     const watcher = watch([], {
         ignoreInitial: true
     }).on('change', async path => {
--- a/src/extension/index.ts
+++ b/src/extension/index.ts
@@ -75,6 +75,8 @@ export async function activate(context: ExtensionContext) {
     }).on('change', async path => {
         store.logs.removeFirst(log => log._uri === Uri.file(path).toString());
         store.logs.push(...await loadLogs([Uri.file(path)]));
+    }).on('unlink', path => {
+        store.logs.removeFirst(log => log._uri === Uri.file(path).toString());
     });
     disposables.push(new Disposable(() => watcher.close()));
 
--- a/src/extension/index.ts
+++ b/src/extension/index.ts
@@ -78,7 +78,7 @@ export async function activate(context: ExtensionContext) {
     }).on('unlink', path => {
         store.logs.removeFirst(log => log._uri === Uri.file(path).toString());
     });
-    disposables.push(new Disposable(() => watcher.close()));
+    disposables.push(new Disposable(async () => await watcher.close()));
 
     // API
     const api = {
--- a/src/extension/index.ts
+++ b/src/extension/index.ts
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+import { watch } from 'chokidar';
 import { diffChars } from 'diff';
 import { observe } from 'mobx';
 import { CancellationToken, commands, DiagnosticSeverity, Disposable, ExtensionContext, languages, OutputChannel, TextDocument, Uri, window, workspace } from 'vscode';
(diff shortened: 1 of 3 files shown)


src/extension/index.ts
```diff
@@ -67,14 +68,28 @@ export async function activate(context: ExtensionContext) {
         update();
     }
 
+    // For now, we keep file watch functionality limited to the API.
+    // After we're sure it's working well, consider also enabling it for files opened manually.
+    const watcher = watch([], {
+        ignoreInitial: true
```
50Wliu, line 74: Not that it matters since we start out with an empty array, but `ignoreInitial` ensures that the change event doesn't get called when the watcher does the first pass on all the files to ready itself.

50Wliu:
> how often do you anticipate quick successions of edits/changes to the same file?

For our use case - never. I can see how difficult it'd be to add a debounce timeout for each file though.

> Note: I did a quick test and Chokidar seems to ignore adding files that have already been added.

Thanks!

> Stylistically

This makes sense. I was thinking it'd be nice to rely on the `initial` functionality. If I get what you're saying - all the API methods would act on `watchedLogs`, which internally would update `logs`. Everything else continues to use `logs` as usual.

Thanks for the quick turnaround! I'll work on those changes.

jeffersonking:
> For our use case - never. I can see how difficult it'd be to add a debounce timeout for each file though.

Glad that's the case. Prefer to defer the debounce until absolutely needed.

chrishuynhc:
Hi @jeffersonking, have you had the chance to finish reviewing? @50Wliu is OOF and asked me to follow up on this change :)

jeffersonking:
@chrishuynhc Not yet. Will do today. Thanks.

jeffersonking:
@50Wliu Good observations and I agree.

Ideally we would be able to add/remove watched logs (say from other code paths) without going the API code path. However, we can do that refactoring when that day comes.

src/extension/index.ts
```diff
@@ -67,19 +68,34 @@ export async function activate(context: ExtensionContext) {
...
+        store.logs.push(...await loadLogs([Uri.file(path)]));
+    }).on('unlink', path => {
+        store.logs.removeFirst(log => log._uri === Uri.file(path).toString());
+    });
+    disposables.push(new Disposable(async () => await watcher.close()));
+
     // API
     const api = {
         async openLogs(logs: Uri[], _options: unknown, cancellationToken?: CancellationToken) {
+            watcher.add(logs.map(log => log.fsPath));
```
jeffersonking, line 86: Should `watcher.add` happen after `store.logs.push`?

jeffersonking (approved):
Approved, pending the remaining comments.

src/extension/index.ts
```diff
@@ -184,13 +200,13 @@ function activateVirtualDocuments(disposables: Disposable[], store: Store) {
             if (contents?.rendered?.text) return contents?.rendered?.text;
             if (contents?.text) return contents?.text;
             if (contents?.binary) {
-                const lines = Buffer.from(contents?.binary, 'base64').toString('hex').match(/.{1,32}/g) ?? [];
-                return lines.reduce((sum, line, i) => {
+                const lines = Buffer.from(contents?.binary, 'base64').toString('hex').match(/.{1,32}/g);
+                return lines?.reduce((sum, line, i) => {
```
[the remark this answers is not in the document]
50Wliu: My local TypeScript version was throwing red errors on this line. The changes aren't required, but I figure they at some point will be when this repo's TS is upgraded. Up to you if you want me to revert the changes, I'm fine either way.

src/extension/index.ts
50Wliu: Hmm. I think that makes sense in case the following happens:
1. We open a log
2. The log is added to the watcher
3. For whatever reason, loadLogs takes forever to finish
4. The log is changed on the filesystem after the watcher starts watching the file but before loadLogs finishes
5. Race condition?

Wouldn't have thought about this otherwise, thanks for pointing it out :)

50Wliu: Actually, turns out that loadLogs isn't even async, so we don't need to worry about that!
...which is a good thing, because my comment made me realize the same race conditions applied to multiple "change" events happening in short succession.

src/extension/index.ts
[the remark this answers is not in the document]
jeffersonking: What was the TS error saying? I'm not seeing the same.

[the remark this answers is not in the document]
50Wliu:
```
This expression is not callable.
  Each member of the union type '{ (callbackfn: (previousValue: string, currentValue: string, currentIndex: number, array: string[]) => string): string; (callbackfn: (previousValue: string, currentValue: string, currentIndex: number, array: string[]) => string, initialValue: string): string; <U>(callbackfn: (previousValue: U, currentValue: string, ...' has signatures, but none of those signatures are compatible with each other.ts(2349)
```
image

Which, after looking it up, seems to be because lines has a type of `RegExpMatchArray | []`, but `.reduce` has to act on a specific type of array.

src/extension/index.ts
jeffersonking: Yes, `loadLogs` is labeled as `async` but doesn't actually `await` anything. In the past we used to await VS Code prompts.

However, if two "change" events come in short succession, it is still possible to race, correct? There's nothing preventing two `loadLogs` calls running at the same time.

src/extension/index.ts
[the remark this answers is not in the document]
jeffersonking: That's a good reason. Let's keep your change.

src/extension/index.ts
50Wliu: Right. Let me think of that and come back to you with a few ideas.

src/extension/loadLogs.ts
```diff
@@ -11,7 +11,7 @@ import * as Telemetry from './telemetry';
 
 export const driverlessRules = new Map<string, ReportingDescriptor>();
 
-export async function loadLogs(uris: Uri[], token?: { isCancellationRequested: boolean }) {
+export function loadLogs(uris: Uri[], token?: { isCancellationRequested: boolean }): Log[] {
```
jeffersonking, line 14: Prefer to defer housekeeping to another PR. There are other calls to `loadLogs` you have not fixed up consistently. But to fix all of them detracts from the clarity of this PR.

src/extension/index.ts
jeffersonking: Since we discussed don't anticipate short succession changes, I'd say a code comment acknowledging the possibility of a race is sufficient.

50Wliu: Sounds good, comment added. Thinking about it - I think if loadLogs is sync, the chance the race condition will actually occur will be extremely small if not impossible - you're banking on another change event getting fired _after_ the removal but _before_ the push.
I'm not familiar with JavaScript multithreading as I'd like to be, but if the extension can only process one "change" event at a time like I suspect - that makes it race-free.

50Wliu:
@jeffersonking, @EasyRhinoMSFT - any chance I could get a signoff + merge? I believe all concerns have been addressed :).

50Wliu:
Thank you! ❤️

EasyRhinoMSFT:
Thank *you* Winston! This is an important feature and you've helped our goal of keeping the two Viewer extensions in sync. 😎

50Wliu:
@EasyRhinoMSFT - would you mind publishing a new release for this, please?

EasyRhinoMSFT:
Sure thing -- apologies for forgetting this, I was on a two-week vacation.

50Wliu:
Thanks so much!

EasyRhinoMSFT:
@50Wliu release to the Marketplace complete 👍

PR #478, opened 2022-12-13
Fix: Hide webview context menu commands as needed.

### LINKED ISSUE #474: "Dismiss" right-click commands are showing up on non-GitHub repos
From [the code](https://github.com/microsoft/sarif-vscode-extension/blob/7bd21f58079a6b35ccdba51a491f2362c204a165/src/extension/index.activateFixes.ts#L34), it seems like these should only show up for GitHub repos, yet they're also showing up on ADO repos for me.
Dismiss actions

Our SARIF does not have github/alertNumber anywhere in it.
No alertNumber

jeffersonking:
Fixes #474.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "sarif-viewer",
-    "version": "3.3.3",
+    "version": "3.3.4",
     "lockfileVersion": 2,
     "requires": true,
     "packages": {
         "": {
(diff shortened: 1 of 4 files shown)


EasyRhinoMSFT (approved):

jeffersonking:
Abandoned due to a GPG signing mistake. Will redo.

PR #484, opened 2023-01-30
Update badge URLs in README.md

EasyRhinoMSFT:
SVGs are only allowed from a whitelisted domain. See https://code.visualstudio.com/api/references/extension-manifest#approved-badges


shaopeng-gh (approved):
:shipit:

PR #495, opened 2023-04-06
Add selectResult function

EasyRhinoMSFT:


src/extension/panel.ts
```diff
@@ -211,6 +211,14 @@ export class Panel {
         this.panel?.webview.postMessage({ command: 'select', id: result?._id });
     }
 
+    public selectByIndex(uri: Uri, runIndex: number, resultIndex: number) {
+        const log = this.store.logs.find(log => log._uri === uri.toString());
+        const result = log?.runs[runIndex]?.results?.[resultIndex];
```
shaopeng-gh, line 216: >. [](http://example.com/codeflow?start=52&length=1)

is this a array do we need the . here

src/extension/panel.ts
```diff
@@ -211,6 +211,14 @@ export class Panel {
         this.panel?.webview.postMessage({ command: 'select', id: result?._id });
     }
 
+    public selectByIndex(uri: Uri, runIndex: number, resultIndex: number) {
```
shaopeng-gh, line 214: >selectByIndex [](http://example.com/codeflow?start=11&length=13)

where is / will it be used?

shaopeng-gh (approved):
:shipit:

src/extension/panel.ts
EasyRhinoMSFT: Will be called by the WAVE extension

src/extension/panel.ts
EasyRhinoMSFT: Good eye

PR #498, opened 2023-04-12
Add selectByIndex to API + fixes

EasyRhinoMSFT:


src/extension/panel.ts
```diff
@@ -213,7 +213,7 @@ export class Panel {
 
     public selectByIndex(uri: Uri, runIndex: number, resultIndex: number) {
         const log = this.store.logs.find(log => log._uri === uri.toString());
-        const result = log?.runs[runIndex]?.results?[resultIndex];
+        const result = log?.runs?.[runIndex]?.results?.[resultIndex];
```
shaopeng-gh, line 216: >?. [](http://example.com/codeflow?start=32&length=2)

I just learn this in Typescript
that we will also need the "." if it is nullable.

shaopeng-gh (approved):
:shipit:

src/extension/panel.ts
```diff
@@ -213,7 +213,7 @@ export class Panel {
 
     public selectByIndex(uri: Uri, runIndex: number, resultIndex: number) {
```
50Wliu, line 214: The implementation of this method could be moved directly into the API (since that's the only caller), but I don't know if that's how you want to organize it or not.

50Wliu (approved):

PR #508, opened 2023-06-06
Honor extension-provided uriBase

### LINKED ISSUE #506: Extension-provided uriBase is ignored
The SARIF VS Code extension provides an extensibility point for other extensions to add their own `uriBases` when resolving SARIF artifact locations:
https://github.com/microsoft/sarif-vscode-extension/blob/main/src/extension/index.d.ts#L15

However, in practice, this never gets used because the extension unconditionally prepends the workspace URI to the artifact location before the code to prepend the custom URI base is ever reached.
https://github.com/microsoft/sarif-vscode-extension/blob/main/src/shared/index.ts#L204

That is to say, if I have a relative SARIF location `my/file/path` and I provide a URI base of `/absolute/base/` while in the `/absolute/base/repo` repo:
Expected: SARIF location resolves to `/absolute/base/my/file/path`
Actual: Extension tries to look up `/absolute/base/repo/my/file/path` as well as `/absolute/base/absolute/base/repo/my/file/path` and fails to locate the file

50Wliu:
Fixes #506

This is a bigger change than I thought it would be, so I would really appreciate some eyes and feedback on it to ensure:
1. This is spec-compliant
2. I'm not breaking existing behavior

The crux of the change is removing all URI parsing logic from `parseArtifactLocation` and instead updating `UriRebaser.translateArtifactToLocal` to handle raw SARIF artifact paths.

Since translateArtifactToLocal now acts on the verbatim artifact URIs, that allows us to easily prepend extension-provided uriBases. However, as a side effect, `artifactUri` is no longer guaranteed to be an actual URI, so lots of the other changes were updating supporting codepaths (e.g. all the base translation) to handle simple filesystem paths, not just URIs.

Validation:
* Relative path and an extension-provided uriBase ✅
* Relative path and a uriBase defined within the SARIF file ✅
* Absolute path that exists on disk ✅
* Relative path that has a distinct project item match ✅
* Relative path that needs location lookup ✅
* Relative path with workspace concatenation ✅
* Relative path that has an open doc match ✅
* Absolute path that doesn't exist on disk but does have a distinct project item match ✅

Sidenotes.
* Since we're operating on FS paths, seems like we should be using `Uri.fsPath` instead of `Uri.path` everywhere, but that's even more potential to breakage so I didn't touch that. Something to keep in mind though.
* It would be nice if `artifactUri` could be renamed to something that didn't imply it was `Uri`, because it isn't (whereas `localUri` _is_ a `Uri`)
* I couldn't run the test suite because it's failing to initialize due to a myriad of pre-existing TypeScript errors.

### DIFF
--- a/src/extension/uriRebaser.ts
+++ b/src/extension/uriRebaser.ts
@@ -12,17 +12,21 @@ import * as fs from 'fs';
 import * as os from 'os';
 import fetch from 'node-fetch';
 
-const workspaceDistinctFilenameCache: Map<string, string | undefined> = new Map();
+const workspaceDistinctFilenameCache: Map<string, Uri | undefined> = new Map();
 
-async function workspaceHasDistinctFilename(filename: string): Promise<string | undefined> {
-    if (workspaceDistinctFilenameCache.has(filename)) {
-        return workspaceDistinctFilenameCache.get(filename);
+async function workspaceHasDistinctFilename(filename: string): Promise<Uri | undefined> {
+    const distinctFileName = workspaceDistinctFilenameCache.get(filename);
+    if (distinctFileName !== undefined) {
+        return distinctFileName;
     }
-    const matches = await workspace.findFiles(`**/${filename}`); // Is `.git` folder excluded?
-    const result = matches.length === 1 ? matches[0].toString() : undefined;
 
-    workspaceDistinctFilenameCache.set(filename, result);
-    return result;
+    const matches = await workspace.findFiles(`**/${filename}`); // Is `.git` folder excluded?
+    if (matches.length === 1) {
+        workspaceDistinctFilenameCache.set(filename, matches[0]);
+        return matches[0];
+    }
+
+    return undefined;
 }
 
 workspace.onDidCreateFiles(async (event) => {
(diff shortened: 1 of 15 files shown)


src/extension/uriRebaser.ts
```diff
@@ -88,10 +86,10 @@ export class UriRebaser {
     // Notes:
     // If 2 logs have the same uri, then likely the same (unless the uri is super short)
     // If 2 logs don't have the same uri, they can still potentially be the same match
-    public async translateLocalToArtifact(localUri: string): Promise<string> { // Future: Ret undefined when certain.
+    public async translateLocalToArtifact(localUri: Uri): Promise<string | undefined> { // Future: Ret undefined when certain.
         // Need to refresh on uri map update.
         if (!this.validatedUrisLocalToArtifact.has(localUri)) {
-            const { file } = platformUriNormalize(localUri);
+            const { file } = platformUriNormalize(localUri).toString();
```
50Wliu, line 92: What is this doing? It lowercases all paths on Windows before passing it into `workspaceHasDistinctFilename`, but is that what we want? Given `workspaceHasDistinctFilename` is actually case-sensitive, even on Windows (https://github.com/microsoft/vscode/issues/10633).

src/extension/uriRebaser.ts
```diff
@@ -88,10 +86,10 @@ export class UriRebaser {
     // Notes:
     // If 2 logs have the same uri, then likely the same (unless the uri is super short)
     // If 2 logs don't have the same uri, they can still potentially be the same match
-    public async translateLocalToArtifact(localUri: string): Promise<string> { // Future: Ret undefined when certain.
+    public async translateLocalToArtifact(localUri: Uri): Promise<string | undefined> {
```
50Wliu, line 89: This can return undefined now - something to check. I can add back a `localUri.fsPath` fallback if that sounds right.

PR #509, opened 2023-06-13
Support `uriBaseId` for artifact locations in fixes.

max-schaefer:
They are supported for other artifact locations (e.g., [here](https://github.com/microsoft/sarif-vscode-extension/blob/d161181d4f04e41d3b5eeaa36b2971b7ff6d2af6/src/extension/index.activateDecorations.ts#L77)), so I think it makes sense to treat fixes the same.

### DIFF
--- a/src/extension/index.activateFixes.ts
+++ b/src/extension/index.activateFixes.ts
@@ -46,7 +46,8 @@ export function activateFixes(disposables: Disposable[], store: Pick<Store, 'ana
                 if (fix) {
                     const edit = new WorkspaceEdit();
                     for (const artifactChange of fix.artifactChanges) {
-                        const [uri, _uriContents] = parseArtifactLocation(result, artifactChange.artifactLocation);
+                        const workspaceUri = workspace.workspaceFolders?.[0]?.uri.toString(); // TODO: Handle multiple workspaces.
+                        const [uri, _uriContents] = parseArtifactLocation(result, artifactChange.artifactLocation, workspaceUri);
                         const artifactUri = uri;
                         if (!artifactUri) continue;
 


src/extension/index.activateFixes.ts
```diff
@@ -46,7 +46,8 @@ export function activateFixes(disposables: Disposable[], store: Pick<Store, 'ana
                 if (fix) {
                     const edit = new WorkspaceEdit();
                     for (const artifactChange of fix.artifactChanges) {
-                        const [uri, _uriContents] = parseArtifactLocation(result, artifactChange.artifactLocation);
+                        const workspaceUri = workspace.workspaceFolders?.[0]?.uri.toString(); // TODO: Handle multiple workspaces.
```
aeisenberg, line 49: I wonder if it would be better to explicitly throw an error here when there is more than one workspace folder. The way it is now, I'm a little concerned that this would fail 50% of the time for multi-root workspaces and users would not be able to debug why.

An even better approach would be to iterate through all folders and find the first one where the artifact exists.

Though, I see that this same `TODO` is in multiple places, so maybe it's fine. Though, I think the text should be this:

```suggestion
                        const workspaceUri = workspace.workspaceFolders?.[0]?.uri.toString(); // TODO: Handle multi-root workspaces.
```

aeisenberg (commented):
Thanks for creating the PR. This looks like it will fix most use cases.

src/extension/index.activateFixes.ts
max-schaefer: I 💯  agree that this isn't a complete fix. The goal of this PR is fairly modest: I just wanted to bring the handling of `uriBaseId` in fixes to parity with how it is handled elsewhere. After that, the logical next step would be to (as you suggest) properly support multi-root workspaces everywhere, but that's a much bigger task.

This is also the reason why I copied the comment verbatim (even though it is a little bit inaccurate), so it will be easier to identify this as another place that needs updating once the `uriBaseId` handling is fixed properly. But I don't feel strongly about that, and am very happy to update the comment as suggested if desired.

50Wliu:
Just curious, does #508 also work for you? I do some larger-scale changes in that PR that also touch upon the code changed in this one, so it would be nice to know if I need to make any additional changes to accommodate this use case :).

max-schaefer:
Yes, that also fixes it. So if it gets merged first I'm very happy to close this PR.

EasyRhinoMSFT:
I did approve @50Wliu's change this morning so we should be good.

max-schaefer:
Brilliant, thanks!

PR #511, opened 2023-06-27
Fix compile errors and warnings

aeisenberg:
Here are a few behaviour preserving changes that I made that helps avoid some compiler errors and warnings.

- Updated launch config for mocha (tests are failing, but at least it launches)
- Updated versions in package-lock.json
- Updated `@types/vscode` to latest
- Removed two `onCommand` triggers that are now deprecated
- Added `const vscode: VsCodeApi` to `global.d.ts` to get properly typed references to `vscode` instance in the webview.
- Converted `return` to `return undefined` where appropriate to make compiler happy.
- Also, a handful of other errors that the compiler was complaining about.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -16,9 +16,8 @@
             "name": "Launch Mocha",
             "type": "node",
             "request": "launch",
-            "program": "${workspaceRoot}/node_modules/mocha/bin/mocha",          
-            "args": ["--color", "--inspect-brk=9229"], // "${file}"
-            "port": 9229,
+            "program": "${workspaceRoot}/node_modules/.bin/mocha",
+            "args": ["--color", "--inspect-brk=9229", "${file}"], //
             "skipFiles": [
                 "<node_internals>/**/*.js",
                 "${workspaceFolder}/node_modules/**/*.js",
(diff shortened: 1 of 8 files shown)


aeisenberg:
@EasyRhinoMSFT, would you be able to take a look at this?

.vscode/launch.json
```diff
@@ -16,9 +16,8 @@
             "name": "Launch Mocha",
             "type": "node",
             "request": "launch",
-            "program": "${workspaceRoot}/node_modules/mocha/bin/mocha",          
-            "args": ["--color", "--inspect-brk=9229"], // "${file}"
-            "port": 9229,
+            "program": "${workspaceRoot}/node_modules/.bin/mocha",
+            "args": ["--color", "--inspect-brk=9229", "${file}"], //
```
EasyRhinoMSFT, line 20: >// [](http://example.com/codeflow?start=66&length=2)

Remove?

EasyRhinoMSFT (approved):
:shipit:

.vscode/launch.json
aeisenberg:
```suggestion
            "args": ["--color", "--inspect-brk=9229", "${file}"],
```

aeisenberg:
Thanks for the review!

PR #513, opened 2023-06-30
Add support for ADO Alert uriHandler

EasyRhinoMSFT:
ADO will link to this extension and pass a URL that will be used to download SARIF for an individual Alert object. The extension will save the log to a temp file and load from there.


src/extension/index.ts
```diff
@@ -48,6 +52,65 @@ export async function activate(context: ExtensionContext) {
     const panel = new Panel(context, baser, store);
     disposables.push(commands.registerCommand('sarif.showPanel', () => panel.show()));
 
+    // URI handler
+    window.registerUriHandler({
```
aeisenberg, line 56: This should be added to the `disposables` list.

src/extension/index.ts
```diff
@@ -48,6 +52,65 @@ export async function activate(context: ExtensionContext) {
...
+        async handleUri(uri: vscode.Uri) {
+            if (uri.path === '/alert') {
```
aeisenberg, line 58: This will handle all URIs with `/alert` as its path, I think. So, `http://evil.com/alert` would be accepted. Is there a way to be more specific?

src/extension/index.ts
```diff
@@ -48,6 +52,65 @@ export async function activate(context: ExtensionContext) {
...
+
+                try {
+                    const jsonObject = await response.json();
+                    await fs.promises.writeFile(filePath, jsonObject.value);
+                } catch (error) {
+                    outputChannel.appendLine(`***Exception in loadAlertSarif\n***${error}\n***File path: ${filePath}\n`);
+                }
+
+                // Load the log into the Viewer.
+                store.logs.push(...await loadLogs([Uri.file(filePath)]));
```
aeisenberg, line 100: Should this go inside the `try` block? If there was an error getting the response or writing to disk, then `loadLogs` will probably not be valid either.

aeisenberg (commented):
Drive-by review here while I am starting to understand the code in this repo.

src/extension/index.ts
EasyRhinoMSFT: This will be the vscode:// uri, in this case vscode://MS-SarifVSCode.sarif-viewer/alert?url=...

src/extension/index.ts
EasyRhinoMSFT: Good point.

EasyRhinoMSFT:
@50wliu could you please take a look at this one

src/extension/index.ts
```diff
@@ -48,6 +52,65 @@ export async function activate(context: ExtensionContext) {
...
+            }
+        }
+    }));
+
+    async function loadAlertSarif(url: URL) {
+        try {
+            const session = await vscode.authentication.getSession('microsoft', ['499b84ac-1321-427f-aa17-267ca6975798/.default'], { createIfNone: true });
+            const accessToken = session?.accessToken;
+
+            const response = await fetch(url, {
```
aeisenberg, line 79: Is there a potential vulnerability here? `url` is coming from user controlled data. So, it could be pointing to any host. This could be a way to steal microsoft access tokens. Should the URL be validated before calling fetch?

src/extension/index.ts
```diff
@@ -48,6 +52,65 @@ export async function activate(context: ExtensionContext) {
...
+                headers: {
+                    'Accept': 'application/sarif+json',
+                    'Authorization': `Bearer ${accessToken}`,
+                    'User-Agent': 'MS-SarifVSCode.sarif-viewer',
+                }
+            });
+
+            if (response.ok) {
+                // Save the log to a new temp file.
+                const filePath = `${os.tmpdir}\\${(new Date()).getTime()}.sarif`;
```
aeisenberg, line 90: Maybe use `path.join` here to be more cross-platform friendly:

```suggestion
                const filePath = path.join(os.tmpdir, (new Date()).getTime()).sarif);
```

src/extension/index.ts
```diff
@@ -63,8 +64,11 @@ export async function activate(context: ExtensionContext) {
                     const pair = param.split('=');
                     if (pair[0] === 'url') {
                         // Decode the alert API URL and pass it to the load function.
-                        await loadAlertSarif(new URL(decodeURIComponent(pair[1])));
-                        break;
+                        const url = decodeURIComponent(pair[1]);
+                        if (url.startsWith('https://advsec.dev.azure.com/')) {
```
aeisenberg, line 68: If the url is incorrect, should there be an error thrown?

EasyRhinoMSFT: Sure, any suggestions on the message? It will be shown on the banner in the lower right corner. I'm not sure how much detail to provide, e.g. what the host must be, etc.

aeisenberg: How about something like:

> `Invalid URL '${url}'. Expecting a URL that starts with 'https://advsec.dev.azure.com/'.`

I am not sure how much control a user has over the URL here. If the URL is invalid, does this mean that there is an invalid alert?

EasyRhinoMSFT: Nope, just means the one they clicked on passed a URL that doesn't point to advsec.

aeisenberg: Ok. Makes sense. I still think a message would be useful so the user knows why the link wasn't opened.

src/extension/index.ts
```diff
@@ -48,6 +53,68 @@ export async function activate(context: ExtensionContext) {
     const panel = new Panel(context, baser, store);
     disposables.push(commands.registerCommand('sarif.showPanel', () => panel.show()));
 
+    // URI handler
+    disposables.push(window.registerUriHandler({
+        async handleUri(uri: vscode.Uri) {
+            if (uri.path === '/alert') {
+                // Launched by Azure DevOps Advanced Security alert page.
+                const pairs = uri.query.split('&');
```
50Wliu, line 61: Consider getting a strongly-typed object via `new URLSearchParams(uri.query)`, which does all the splitting work for you.

src/extension/index.ts
```diff
@@ -20,6 +21,10 @@ import { Store } from './store';
 import * as Telemetry from './telemetry';
 import { update, updateChannelConfigSection } from './update';
 import { UriRebaser } from './uriRebaser';
+import * as fs from 'fs';
```
50Wliu, line 24: You can act directly on the `promises` object by importing from `'fs/promises'` instead

package.json
```diff
@@ -188,6 +188,11 @@
         "webpack-dev-server": "^4.2.1"
     },
     "dependencies": {
+        "@azure/arm-subscriptions": "^5.1.0",
```
50Wliu, line 191: None of these dependencies seem to be used - am I missing something?

src/extension/index.ts
```diff
@@ -48,6 +53,68 @@ export async function activate(context: ExtensionContext) {
...
+                    outputChannel.appendLine(`***Exception in loadAlertSarif\n***${error}\n***File path: ${filePath}\n`);
+                }
+            } else {
+                outputChannel.appendLine(`***Failed request in loadAlertSarif\n***Response code ${response.status} ${response.statusText}\n***URL: ${url.toString()}\n`);
+            }
+        } catch (error) {
+            outputChannel.appendLine(`***Exception in loadAlertSarif\n***${error}\n***URL: ${url.toString()}\n`);
+        }
+
+        return undefined;
```
50Wliu, line 113:
```suggestion
```

src/extension/index.ts
```diff
@@ -48,6 +53,68 @@ export async function activate(context: ExtensionContext) {
...
+                    }
+                }
+            }
+        }
+    }));
+
+    async function loadAlertSarif(url: URL) {
+        try {
+            const session = await vscode.authentication.getSession('microsoft', ['499b84ac-1321-427f-aa17-267ca6975798/.default'], { createIfNone: true });
+            const accessToken = session?.accessToken;
```
50Wliu, line 81: Should probably exit early here if they didn't sign in, rather than sending a useless HTTP request without authorization.

src/extension/index.ts
```diff
@@ -58,17 +58,19 @@ export async function activate(context: ExtensionContext) {
...
+                for (const param of params) {
+                    if (param[0] === 'url') {
                         // Decode the alert API URL and pass it to the load function.
-                        const url = decodeURIComponent(pair[1]);
+                        const url = decodeURIComponent(param[1]);
                         if (url.startsWith('https://advsec.dev.azure.com/')) {
                             await loadAlertSarif(new URL(url));
-                            break;
+                        } else {
+                            await window.showWarningMessage(`Invalid callback URL '${url}'. URL must point to 'https://advsec.dev.azure.com/'.`, 'OK');
```
aeisenberg, line 71: There needs to be a `throw` or `return` here, otherwise the call to `fetch(url)` will happen and the access token will be sent to an arbitrary url. 

Also, probably you want to change `await` to `void` since we most likely don't want to stop execution while waiting for the user to close the dialog.

EasyRhinoMSFT: Not sure on the first point. If the param is named 'url', it breaks out of the for loop and that's the end of the function.

aeisenberg: I'm sorry. You're right. The function finishes. Please ignore.

aeisenberg (approved):
This looks good to me, but you may want to have someone more familiar with the code to do a second approval.

src/extension/index.ts
```diff
@@ -48,6 +53,68 @@ export async function activate(context: ExtensionContext) {
     const panel = new Panel(context, baser, store);
     disposables.push(commands.registerCommand('sarif.showPanel', () => panel.show()));
 
+    // URI handler
+    disposables.push(window.registerUriHandler({
+        async handleUri(uri: vscode.Uri) {
+            if (uri.path === '/alert') {
```
shaopeng-gh, line 59: >'/alert' [](http://example.com/codeflow?start=29&length=8)

just a note if we need case sensitive

src/extension/index.ts
```diff
@@ -48,6 +53,68 @@ export async function activate(context: ExtensionContext) {
...
+                        break;
+                    }
+                }
+            }
+        }
+    }));
+
+    async function loadAlertSarif(url: URL) {
+        try {
+            const session = await vscode.authentication.getSession('microsoft', ['499b84ac-1321-427f-aa17-267ca6975798/.default'], { createIfNone: true });
```
shaopeng-gh, line 82: >499b84ac-1321-427f-aa17-267ca6975798 [](http://example.com/codeflow?start=82&length=36)

do we need a comment what this guid is for

src/extension/index.ts
shaopeng-gh: hey I found below see if it fits:

```
if (uri.scheme === 'myextension') {
        if (uri.path === '/open-file') {
    }
}
```

shaopeng-gh (approved):
:shipit:

Issue #515, opened 2023-07-25
Extension fails to load analyses if git repo has no `origin` remote


aeisenberg:
If your repo does not use `origin` as the remote name, then the extension will fail to download any analyses.

A PR is forthcoming where if the `origin` does not exist, the first remote listed in the config file will be used instead. And if no remotes exist, then the method exits and issues proper telemetry.

PR #516, opened 2023-07-25
Avoid crashing if a git repo has no `origin`

aeisenberg:
If there is no `origin`, then choose the first remote in the config. If there are no remotes, exit early and send telemetry.

Fixes https://github.com/microsoft/sarif-vscode-extension/issues/515

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -7,7 +7,6 @@
             "request": "launch",
             "runtimeExecutable": "${execPath}",
             "args": ["--extensionDevelopmentPath=${workspaceRoot}"],
-            "stopOnEntry": false,
             "sourceMaps": true,
             "outFiles": [ "${workspaceRoot}/out/**/*.js" ],
             "preLaunchTask": "npm: start"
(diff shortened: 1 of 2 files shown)


50Wliu:
This approach looks like it should work, but you could also consider how the Azure Pipelines extension does it, by looking at HEAD's remote:
https://github.com/microsoft/azure-pipelines-vscode/blob/914ccb70757d5c07bf563bac2ce41ba6fa208d7d/src/schema-association-service.ts#L134-L139
(or, alternatively, grabbing the first remote from `repo.state.remotes`)

aeisenberg:
Thanks for the suggestion. So, this works if the current branch is connected to an upstream remote? It probably makes more sense to try this first rather than looking for `origin`.

.vscode/launch.json
```diff
@@ -7,7 +7,6 @@
             "request": "launch",
             "runtimeExecutable": "${execPath}",
             "args": ["--extensionDevelopmentPath=${workspaceRoot}"],
-            "stopOnEntry": false,
```
aeisenberg, line 10: This property is unused.

50Wliu (approved):
This looks good, but I don't think I'm the right person to merge this.

aeisenberg:
Thanks, @50Wliu.

@EasyRhinoMSFT are you ok with this?

PR #518, opened 2023-08-10
Fix AdvSec URL extraction

EasyRhinoMSFT:
decodeURIComponent doesn't handle the url param correctly. It drops all but the first param because it thinks the ampersands are part of the parent query string. So this change just splits on the param name we expect and uses the whole value as the AdvSec target API.


shaopeng-gh (approved):
:shipit:

PR #521, opened 2023-08-14
Add onUri activation event

EasyRhinoMSFT:
Loads the extension if necessary when a URI targets it

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
     "description": "Adds support for viewing SARIF logs",
     "author": "Microsoft Corporation",
     "license": "MIT",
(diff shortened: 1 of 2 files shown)


50Wliu (approved):

package.json
```diff
@@ -4,7 +4,7 @@
     "description": "Adds support for viewing SARIF logs",
     "author": "Microsoft Corporation",
     "license": "MIT",
-    "version": "3.4.0",
+    "version": "3.4.1",
```
50Wliu, line 7: Might need to regenerate package-lock again to pick up this change

PR #522, opened 2023-08-22
Add functionality to download analysis on startup

aeisenberg:
If the `sarif-viewer.githubCodeScanningAlertId` setting has a value, use the value to download an appropriate code scanning analysis from  github and apply any fixes found in it.


aeisenberg:
@nickfyson, this should work for downloading autofixes on startup, but of course I can't really test this until we have some sample autofix data to play with.

aeisenberg:
This PR is ready for review now. I've confirmed that it is working. However, there may need to be some tweaks as we formalize the feature on the code scanning side.

aeisenberg:
@EasyRhinoMSFT @50Wliu would one of you be able to take a look at this PR? I'm happy to chat about this to explain some of the changes.

aeisenberg:
I just added two more commits. When reviewing this PR, it will be helpful to review commit-by-commit.

src/extension/index.activateFixes.ts
```diff
@@ -98,3 +80,47 @@ class DismissCodeAction extends CodeAction {
...
     }
 }
+
+export async function applyFix(fix: Fix, result: Result, baser: UriRebaser, store: Pick<Store, 'analysisInfo'>) {
+    if (fix.properties?.diff) {
+        const git = await getInitializedGitApi();
+        if (!git) {
+            throw new Error('Unable to initialize Git API.');
+        }
+        const diff = fix.properties?.diff;
```
50Wliu: Diff is known to be non-null here because if the check above...maybe extract it into a variable _before_ the if.

src/extension/index.activateGithubAnalyses.ts
```diff
@@ -156,24 +160,36 @@ export function activateGithubAnalyses(disposables: Disposable[], store: Store,
...
-            await onBranchChanged(repo, gitHeadPath, store);
+
+            if (fullCodeScanningAlert) {
+                // bypass downloading analyses. Instead, use the fullCodeScanningAlert as the analysis to apply.
+                await handleSingleLog(fullCodeScanningAlert);
+
+                // Now that the analysis has been successfully applied, avoid re-applying it in the future.
+                // Make sure to update the config value to '' instead of `undefined` to ensure that the
+                // default value is not used. This default value may have been injected by a codespace.
+                await workspace.getConfiguration('sarif-viewer').update('githubCodeScanningInitialAlert', '', ConfigurationTarget.Global);
```
50Wliu: Given that this seems to be internal state, a better storage choice for this might be `globalState` on the extension's [context](https://code.visualstudio.com/api/references/vscode-api#ExtensionContext).

src/extension/index.activateGithubAnalyses.ts
```diff
@@ -394,6 +441,19 @@ export function activateGithubAnalyses(disposables: Disposable[], store: Store,
...
 
+function errorToString(e: unknown) {
+    return e instanceof Error ? e.message : String(e);
+}
+
+function parseLog(logText: string, uri?: string) {
+    const log = JSON.parse(logText) as Log;
+    log._text = logText;
+    log._uri = uri ?? 'file:///synthetic.sarif';
+    const primaryWorkspaceFolderUriString = workspace.workspaceFolders?.[0]?.uri.toString(); // No trailing slash
```
50Wliu, line 480: You might want `toString(true)`...I think.

src/extension/index.activateGithubAnalyses.ts
```diff
@@ -394,6 +441,19 @@ export function activateGithubAnalyses(disposables: Disposable[], store: Store,
     });
 }
 
+function errorToString(e: unknown) {
+    return e instanceof Error ? e.message : String(e);
+}
+
+function parseLog(logText: string, uri?: string) {
```
50Wliu, line 476: Give uri the default value of `file:///synthetic.sarif` here

src/extension/index.activateFixes.ts
```diff
@@ -98,3 +80,47 @@ class DismissCodeAction extends CodeAction {
...
+        // save diff to a temp file
+        const filePath = path.join(os.tmpdir(), `${(new Date()).getTime()}.patch`);
+
+        try {
+            await workspace.fs.writeFile(Uri.parse(filePath), Buffer.from(diff, 'utf-8'));
+
+            // TODO assume exactly one repository, which will be the case when working in a codespace.
+            await git?.repositories[0].apply(filePath);
+        } finally {
+            await workspace.fs.delete(Uri.parse(filePath));
```
50Wliu, line 104: Telemetry on failure to apply patch?

src/extension/index.activateGithubAnalyses.ts
```diff
@@ -468,16 +469,28 @@ function parseLog(logText: string, uri?: string) {
...
  * @param repo The repo to retrieve analyses for
  * @returns the url associated with this remote or undefined if there are no remotes
  * for this repo.
  */
-function findRemote(repo: Repository) {
-    const remoteName = repo.state.HEAD?.upstream?.remote || 'origin';
-    let remoteUrl = repo.state.remotes.find(remote => remote.name === remoteName)?.fetchUrl;
-    if (!remoteUrl && repo.state.remotes.length) {
-        remoteUrl = repo.state.remotes[0].fetchUrl;
+async function findRemote(repo: Repository, outputChannel: OutputChannel): Promise<string | undefined> {
```
50Wliu, line 502: FWIW, I worked around this in the Azure Pipelines extension by calling `await openRepository(the_repo)` rather than `getRepository(the_repo)`, and it seems to have solved the issue.

(in my opinion - bad API design on the Git side for not handling the startup case for us.)

src/extension/index.activateFixes.ts
```diff
@@ -81,26 +81,31 @@ class DismissCodeAction extends CodeAction {
...
-
             // TODO assume exactly one repository, which will be the case when working in a codespace.
             await git?.repositories[0].apply(filePath);
+            outputChannel?.appendLine('diff applied.');
         } finally {
             await workspace.fs.delete(Uri.parse(filePath));
         }
         return;
     }
+    outputChannel?.appendLine(JSON.stringify('Edit found.'));
```
50Wliu, line 108: I don't think this JSON.stringify is needed 😁

src/extension/index.activateFixes.ts
```diff
@@ -98,3 +80,53 @@ class DismissCodeAction extends CodeAction {
...
+        outputChannel?.appendLine('--------');
+        const git = await getInitializedGitApi();
+        if (!git) {
+            throw new Error('Unable to initialize Git API.');
+        }
+        // save diff to a temp file
+        const filePath = path.join(os.tmpdir(), `${(new Date()).getTime()}.patch`);
+        try {
+            await workspace.fs.writeFile(Uri.parse(filePath), Buffer.from(diff, 'utf-8'));
+            // TODO assume exactly one repository, which will be the case when working in a codespace.
```
50Wliu, line 100: I have added multiple repos to my workspace in codespaces before 😉. The authentication didn't exactly work...but it _is_ possible to have multi-repo workspaces.

50Wliu (approved):
Approving in case you want to get this out, but I think the new config option makes more sense as workspace/global state.

src/extension/index.activateGithubAnalyses.ts
aeisenberg: Unfortunately, it can't be global state. This needs to be injected as a setting by codespaces when the vm is starting up. There is no way to use global state for that AFAIK.

src/extension/index.activateGithubAnalyses.ts
aeisenberg: Oh..I didn't know about that API. I will look into it.

src/extension/index.activateFixes.ts
aeisenberg: You're right. There can be multiple repositories. In all the cases we need to handle right now, there will only be a single repository. I'll change the comment and we may need to revisit in the future.

src/extension/index.activateFixes.ts
aeisenberg: This information is handled higher up the call stack in `handleSingleLog`.

src/extension/index.activateGithubAnalyses.ts
aeisenberg: On a deeper look, it seems like using `openRepository` would be a bigger change and touch several areas outside the current PR. I'll create an issue for this to work on in the future, but I'd rather not bring unrelated changes in to this PR.

src/extension/index.activateGithubAnalyses.ts
50Wliu: Ah, unfortunate. All right, then.

50Wliu (approved):

PR #523, opened 2023-08-24
Fixes unit tests

aeisenberg:
With this change, running `npm run tests` will succeed.

It looks like the tests have not been run for a long time and the implementation has drifted, making the tests harder to run.

There are lots of small changes here:

1. Change the `launch.json` to remove `--inspect-brk=9229`, which was preventing the tests from running.
2. Add some exclusions to `settings.json` so that some unnecessary folders are removed from the file open dialog.
3. Fix up the mock objects so that the tests correctly reflect the current implementation.
4. Add `await` and `async` to `setState` in indexStore.ts so that the function correctly waits for the post message to complete. (Note this is the _only_ change to non-test code in this PR.)
5. Change the test assertions so that they test the right things.

A few comments here on future work:

1. These tests all use stubs for the vscode API. This isn't ideal since it's complicated and we really should be testing the vscode API directly. Future work should be to run the unit tests from inside of a workbench so we don't need to stub out the vscode API.
2. We should be running the tests on each PR. I'll fix this in a followup PR.

To validate this PR, please run `npm run tests` locally and ensure the tests pass for you.

### DIFF
--- a/src/extension/index.spec.ts
+++ b/src/extension/index.spec.ts
@@ -20,7 +20,6 @@ describe('activate', () => {
     before(async () => {
         const { activate } = proxyquire('.', {
             'fs': {
-                // '@global': true,
                 readFileSync: () => {
                     return mockLogString;
                 }
--- a/src/extension/index.spec.ts
+++ b/src/extension/index.spec.ts
@@ -6,10 +6,10 @@
 /// Todo: Migrate to tsconfig.files
 
 import assert from 'assert';
-import { URI as Uri } from 'vscode-uri';
 import { postSelectLog } from '../panel/indexStore';
 import { log } from '../test/mockLog';
-import { mockVscode, mockVscodeTestFacing } from '../test/mockVscode';
+import { mockVscode, mockVscodeTestFacing, uriForRealFile } from '../test/mockVscode';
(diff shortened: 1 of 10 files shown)


aeisenberg:
@EasyRhinoMSFT could you take a look at this? I'd like to eventually run the unit tests on every PR.

src/extension/index.spec.ts
```diff
@@ -20,7 +20,7 @@ describe('activate', () => {
     before(async () => {
         const { activate } = proxyquire('.', {
             'fs': {
-                '@global': true,
+                // '@global': true,
```
EasyRhinoMSFT, line 23: >// '@global': true, [](http://example.com/codeflow?start=16&length=19)

Do we need to keep this? (I saw others too)

aeisenberg: Right...I can remove this one. I'll see if I can remove the others. Best if we can avoid `@global`.

aeisenberg: Just verified...it needs to stick around for the `vscode` imports. All others can be removed.

aeisenberg:
Thanks for the review!

PR #524, opened 2023-09-01
Run tests during CI

aeisenberg:
This change renames `eslint.yml` to `checks.yml` and adds a new job that runs the tests in that workflow.

There are some additional changes associated with this:

1. The tests in `index.spec.ts` were hanging after they ran since they did not properly dispose the API after running. This is now fixed.
2. I added the `dispose` method to the `api`, which properly disposes all required objects.
3. I removed the `deactivate` method from `index.ts` since it is not used anywhere.
4. Similarly, I deleted `_options` from `openLogs` since it is not used.
5. I removed `async` from several api methods that do not need it.
6. I updated the typing in `index.d.ts` to properly reflect the type of the api.


src/extension/index.d.ts
```diff
@@ -9,8 +9,9 @@ export interface Api {
      * @param logs An array of Uris to open.
      */
     openLogs(logs: Uri[]): Promise<void>;
-    closeLogs(logs: Uri[]): Promise<void>;
-    closeAllLogs(): Promise<void>;
-    selectByIndex(uri: Uri, runIndex: number, resultIndex: number): Promise<void>;
+    closeLogs(logs: Uri[]): void;
```
50Wliu, line 12: This is the public API used by other extensions integrating with the SARIF Viewer - and as such might be a breaking change. I think if extensions are using `await closeAllLogs()`, they'll be fine, but if they're doing `closeAllLogs().then(/* ... */)` that will start breaking.

src/extension/index.ts
```diff
@@ -151,23 +152,27 @@ export async function activate(context: ExtensionContext) {
 
     // API
     const api = {
-        async openLogs(logs: Uri[], _options: unknown, cancellationToken?: CancellationToken) {
+        async openLogs(logs: Uri[], cancellationToken?: CancellationToken) {
```
50Wliu, line 155: Also a breaking change

50Wliu (changes requested):
If you go ahead with the Promise changes, I think you'd need to create a new API version to avoid breaking existing consumers (like us!).

src/extension/index.d.ts
aeisenberg: Ah...thanks for letting me know. I didn't realize that this was a public facing API used by other extensions. I can undo these changes to ensure compatibility. I'll also add a comment so that we don't make any accidental changes in the future.

src/extension/index.ts
aeisenberg: Should this be added to the .d.ts file?

50Wliu: We would appreciate for it to be (we do pass in a cancellation token), but Chris probably has the most context on whether it _should_ be.

src/extension/index.spec.ts
```diff
@@ -10,13 +10,17 @@ import { postSelectLog } from '../panel/indexStore';
...
 
 // Log object may be modified during testing, thus we need to keep a clean string copy.
 const mockLogString = JSON.stringify(log, null, 2);
 
 const proxyquire = require('proxyquire').noCallThru();
 
-describe('activate', () => {
+let api: Api;
+
+// TODO Tests are hanging on CI.
```
50Wliu, line 22: How could you tell this suite is the one that's hanging from the logs? Nothing really stood out to me.

50Wliu (approved):
Rubberstamp approval to clear my previous rejection.

src/extension/index.spec.ts
aeisenberg: It was an educated guess since this was the suite that was hanging locally for me until I added the correct `dispose()` method and called it from the tests.

I'm guessing that there is something that is not being disposed properly on linux. Possibly, this is something about file watchers (since this is os-dependent and it is passing for me on MacOS). Figuring this out will take a lot of trial and error and be very time consuming.

I'll update the comment with some of my thoughts on what might be happening so it will be easier to pick up in the future.

PR #526, opened 2023-09-22
Convert to linux runner for creating releases

aeisenberg:
This will make it faster to run this workflow. As it is, getting a macos runner can take a while.


EasyRhinoMSFT (approved):
:shipit:

Issue #548, opened 2024-02-06
100% cpu usage 


cosdon:
Type: <b>Bug</b>

Open a large workspace with many node_modules directories; typing produces 100% cpu usage due to multiple copies of ripgrep being started.

Extension version: 3.4.2
VS Code version: Code 1.74.3 (97dec172d3256f8ca4bfb2143f3f76b503ca0534, 2023-01-09T17:07:18.579Z)
OS version: Darwin x64 23.2.0
Modes:
Sandboxed: No

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz (16 x 2300)|
|GPU Status|2d_canvas: enabled<br>canvas_oop_rasterization: disabled_off<br>direct_rendering_display_compositor: disabled_off_ok<br>gpu_compositing: enabled<br>metal: disabled_off<br>multiple_raster_threads: enabled_on<br>opengl: enabled_on<br>rasterization: enabled<br>raw_draw: disabled_off_ok<br>skia_renderer: enabled_on<br>video_decode: enabled<br>video_encode: enabled<br>vulkan: disabled_off<br>webgl: enabled<br>webgl2: enabled<br>webgpu: disabled_off|
|Load (avg)|39, 44, 30|
|Memory (System)|32.00GB (0.36GB free)|
|Process Argv|--crash-reporter-id adeab385-da39-4c80-8779-4b6d6d1a8e71|
|Screen Reader|no|
|VM|0%|
</details><details>
<summary>A/B Experiments</summary>

```
vsliv368:30146709
vspor879:30202332
vspor708:30202333
vspor363:30204092
vscod805:30301674
binariesv615:30325510
vsaa593:30376534
py29gd2263:30899288
c4g48928:30535728
azure-dev_surveyone:30548225
2i9eh265:30646982
962ge761:30951796
pythongtdpath:30769146
pythonidxpt:30866567
pythonnoceb:30805159
asynctok:30898717
pythontestfixt:30902429
pythonregdiag2:30936856
pyreplss1:30897532
pythonmypyd1:30879173
pythoncet0:30885854
pythontbext0:30879054
dsvsc016:30899300
dsvsc017:30899301
dsvsc018:30899302
dsvsc019b:30953937
3ef8e399:30949928
e3gdj431:30958358
ccp1r6:30958156

```

</details>

renchap:
[...]

PR #556, opened 2024-03-21
Address CPU usage spike: Remove uriRebaser normalize and added workspace.findFiles maxResults value

chrishuynhc:
## Overview
This change addresses the high CPU usage issue caused by the SARIF Viewer mentioned here: https://github.com/microsoft/sarif-vscode-extension/issues/548

For context, the high CPU usage was caused by the spawning of `rg.exe` (rigrep) processes to perform VSCode's `workspace.findFiles` searches. This file search is triggered on text document changes and is used for the uri rebaser logic of matching workspace files to SARIF log artifacts.

## The Issue/Fix
The issue was due to the `vscode.worspace.findFiles` method being case-sensitive while the path being searched is normalized to lowercase.  So, if the user happens to have a file open that contains capitalization in the file name, such as `fooBar.ts`, the file search would look for `foobar.ts` and `vscode.worspace.findFiles` would result in 0 matches.  As a result, this file search would repeat as fooBar.ts would never be found and cached resulting in repetitive calls to `vscode.worspace.findFiles` when trying to resolve artifact matches and fill the distinct files cache.

To address this, the file uri is no longer normalized (converted to lowercase) when performing the workspace search and cache check. The normalize method was meant to handle the Windows platform (case-insensitive), but because VS Code itself and `vscode.worspace.findFiles` is case-sensitive in addition to the normalize method only being used for this distinct filename flow, we can remove it.

In addition, this issue is compounded on larger workspace which is why a `maxLimit` of 5000 files is now specified when utilizing `vscode.worspace.findFiles.`  This helps to prevent perf issues when working in large workspaces such as the windows OS repo by stopping the search when the limit is hit.


EasyRhinoMSFT (approved):
:shipit:

50Wliu (approved):

stmillMSFT:
This issue is absolutely still repro, btw. I'm having this issue too. Same thing: no SARIF files in repo, and not even that big of a repo, but about a million rg.exe instances are spawned, and I can't type anymore. Since this extension seems to have been pushed out via group policy for some unknown reason, please address as a higher severity; this is making internal work miserable.

PR #562, opened 2024-04-03
Give GITHUB_TOKEN release permissions

50Wliu:
GITHUB_TOKEN was changed to default to read-only earlier this year: https://docs.opensource.microsoft.com/github/apps/permission-changes/

The right permission to add back is write access to [`contents`](https://docs.github.com/en/rest/authentication/permissions-required-for-github-apps?apiVersion=2022-11-28#repository-permissions-for-contents) (which covers...a scary amount of things).

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -6,6 +6,9 @@ jobs:
   createPrerelease:
     name: Create Release
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


aeisenberg (approved):
Looks right to me. Unfortunately, there's no way to restrict the permission to only releases.

PR #584, opened 2025-03-11
feat: add justification to SARIF info table

carsonRadtke:


src/panel/details.tsx
```diff
@@ -50,6 +50,17 @@ interface DetailsProps { result: Result, resultsFixed: string[], height: IObserv
...
         };
 
+        const renderSuppressionInformation = (result: Result) => {
+            const text = result._suppression;
+            const justification = result._justification;
+            if (!text && !justification) return '—'; // unreachable
+            if (!justification)
+            {
+                return text;
+            }
```
aeisenberg, line 60: Nit: formatting
```suggestion
            if (!text && !justification) {
                return '—'; // unreachable
            }
            if (!justification) {
                return text;
            }
```

src/shared/index.ts
```diff
@@ -132,6 +133,15 @@ export function augmentLog(log: Log, rules?: Map<string, ReportingDescriptor>, w
             result._suppression = !result.suppressions || result.suppressions.every(sup => sup.status === 'rejected')
                 ? 'not suppressed'
                 : 'suppressed';
+            result._justification = undefined;
+            if (result.suppressions)
+            {
+                const justifications = result.suppressions.filter(sup => sup && sup.justification).map(sup => sup.justification!);
```
aeisenberg, line 139: I prefer trying to avoid type assertions `!`. I think this should do it.

```suggestion
                const justifications = result.suppressions
                    .map(sup => sup?.justification)
                    .filter(justification => !!justification);
```

aeisenberg (commented):
Looks good. A couple of stylistic comments.

aeisenberg (commented):
I have not tried this out myself. The code looks good, but we don't have that many tests. Please verify that this works for you locally and I will merge.

carsonRadtke:
Here is a warning (not suppressed):
image

Here is a warning (suppressed, no justification):
image

Here is a warning (suppressed, w/ justification):
image

I have also attached the SARIF file for convenience (renamed to .json for GitHub): [foo.nativecodeanalysis.json](https://github.com/user-attachments/files/19199200/foo.nativecodeanalysis.json)

aeisenberg (approved):
Thanks for the screenshots.

PR #607, opened 2025-06-30
Fix: Annotation filter incorrectly showing all files

### LINKED ISSUE #606: bug: finding annotations stay even in after switching to other file
Finding's step annotations stay even  after switching to other file.
This bug also related to an exception thrown:
```
rejected promise not handled within 1 second: Error: Illegal value for `line`
extensionHostProcess.js:177
stack trace: Error: Illegal value for `line`
    at cW.u (file:///c:/Users/Uriya%20Yavnieli/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/workbench/api/node/extensionHostProcess.js:128:1045)
    at Object.lineAt (file:///c:/Users/Uriya%20Yavnieli/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/workbench/api/node/extensionHostProcess.js:128:72)
    at regionToSelection (c:\Users\Uriya Yavnieli\.vscode\extensions\ms-sarifvscode.sarif-viewer-3.4.5\src\extension\regionToSelection.ts:30:26)
    at Object.driftedRegionToSelection (c:\Users\Uriya Yavnieli\.vscode\extensions\ms-sarifvscode.sarif-viewer-3.4.5\src\extension\regionToSelection.ts:55:43)
    at map (c:\Users\Uriya Yavnieli\.vscode\extensions\ms-sarifvscode.sarif-viewer-3.4.5\src\extension\index.activateDecorations.ts:82:54)
    at Array.map (<anonymous>)
    at u (c:\Users\Uriya Yavnieli\.vscode\extensions\ms-sarifvscode.sarif-viewer-3.4.5\src\extension\index.activateDecorations.ts:82:43)
```
I saw that the `update()` in sarif-vscode-extension/src/extension/index.activateDecorations.ts walks on every visible editor and add decorations to it:
```typescript
const ranges = locationsInDoc.map(tfl => driftedRegionToSelection(diffBlocks, currentDoc, tfl.location?.physicalLocation?.region, originalDoc));
editor.setDecorations(decorationTypeHighlight, ranges);
```
the function `driftedRegionToSelection()` calls `regionToSelection()`:
```typescript
const originalRange = regionToSelection(originalDoc, region);
```
which then tries to get the line from range.startLine - 1:
```typescript
const line = doc.lineAt(startLine - 1);
```
Since `regionToSelection()` is called for every visible editor, it will also might be called for text document that are too short, if they are visible, which ends up with the exception from above.
And anyway the decorations are being added to irrelevant documents.

However, the `update()` function do filter locations by the URI:
```typescript
const locationsInDoc = locations.filter(async tfl => {
    const [artifactUriString] = parseArtifactLocation(result, tfl.location?.physicalLocation?.artifactLocation);
    return await baser.translateLocalToArtifact(currentDoc.uri) === artifactUriString;
});
```
But `Array.filter()` doesn't handle async callback properly.
Here's what's happening according to Claude Sonnet 4:
1. The callback function is async, which means it returns a Promise<boolean> instead of a boolean
2. Array.filter() expects a synchronous callback that returns a boolean
3. When filter() receives a Promise, it treats it as a "truthy" value, so ALL items will pass the filter regardless of the actual async result
4. This means locationsInDoc will always contain all items from locations, not just the filtered ones

So all the items will always pass the filter.

The suggested fix according to Claude is:
```typescript
// Properly handle async filtering
const locationChecks = await Promise.all(
    locations.map(async tfl => {
        const [artifactUriString] = parseArtifactLocation(result, tfl.location?.physicalLocation?.artifactLocation);
        const matches = await baser.translateLocalToArtifact(currentDoc.uri) === artifactUriString;
        return { tfl, matches };
    })
);
const locationsInDoc = locationChecks.filter(check => check.matches).map(check => check.tfl);
```

uriyay-jfrog:
fixes https://github.com/microsoft/sarif-vscode-extension/issues/606


joshbouncesecurity:
Thanks for this @uriyay-jfrog, shame this project seems to have been abandoned...

Issue #415, opened 2022-01-19
String `path` monkey patch breaks form-data library


ahamid:
Is it necessary to monkey patch `String` like this:

https://github.com/microsoft/sarif-vscode-extension/blob/bd13e8021549fb1674795a9bd473803318dfe673/src/shared/extension.ts#L75

```javascript
Object.defineProperty(String.prototype, 'path', {
    get: function() {
        return this.substring(0, this.lastIndexOf('/')).replace(/^\//g, '');
    }
});
```

I found, for example, that introducing a `path` property/accessor can break the `form-data` library in a really subtle and non-obvious way:

https://github.com/form-data/form-data/blob/53adbd81e9bde27007b28083068f2fc8272614dc/lib/form_data.js#L105

For anyone who runs into this the solution appears to be to marshal the string value into a `Buffer` with `Buffer.from` to avoid detection of the accessor.

markieparky:
I had the exact same problem, causing my own extension to fail. Adding a global primitive augmentations is a massive anti-pattern and introduces unpredictable side effects across the entire runtime.

Thanks ahamid for the workaround.

PR #622, opened 2025-11-18
Removed global String and Array prototype augmentations to avoid conflicts with other extensions

markieparky:
As per issue: https://github.com/microsoft/sarif-vscode-extension/issues/415

I have removed the global augmentations and changed the functionalities to functions. 
I have tried to adhere to the code style as best as I could, if some code needs to be moved let me know.


markieparky:
@microsoft-github-policy-service agree

PR #627, opened 2025-12-22
Make open panel optional in API

wiomoc:
Provides the preference to open the side panel via options.

Also fixes the signature of both `openLogs` & `closeLogs`.


wiomoc:
@microsoft-github-policy-service agree company="Vector Informatik Gmbh"

Issue #620, opened 2025-11-18
Populate "code" in the VSCode problem panel


yairlenga:
After loading SARIF files, the "code" column of the "problems" panel is showing "-". The most likely relevant value is to use the "results[x].ruleId". Requesting to add this behavior, which I believe will make the viewer better.

Implementation should be simple:
In src/extension/resultDiagnostic.ts - change the construct (which already takes result) to set the 'code' attribute

```
export class resultDiagnostic extends Diagnostic {
    construct(...) {
        super(...)
        this.code = { value = result?.ruleId }
    }
}
```

yairlenga:
Hi folks — quick ping in case this got buried. I sent a PR for this change, Happy to make any adjustments if needed. Thanks!

PR #630, opened 2026-01-11
Expose SARIF ruleId as diagnostic code in Problems view

yairlenga:
References #620 

Expose SARIF result `ruleId` as the VS Code diagnostic code so it appears in:

1. the Problems view (e.g. `SARIF(ruleId)`) in table mode - currently showing '-'
2. In the problem popup. Currently not showing source/code.

This is consistent with the way other extensions (e.g. TS, C/C++) present errors. No behavior change when `ruleId` is not present. 

Sample output before/after - using the current test case (`demos/distinctNameMatching/.sarif/log.sarif`) - labeling the issue as `SARIF(DEMO01)`

Before:
image

After:
image


yairlenga:
Hi folks — quick ping in case this got buried. This PR adds ruleId → code column support and should be ready to merge (tests based on existing test cases included). Happy to make any adjustments if needed. Thanks!
@mention @aeisenberg @chrishuynhc @huskydawg

PR #645, opened 2026-07-31
Initialize primary Git repository before use

### LINKED ISSUE #525: Convert to using `openRepository` API
FWIW, I worked around this in the Azure Pipelines extension by calling `await openRepository(the_repo)` rather than `getRepository(the_repo)`, and it seems to have solved the issue.

(in my opinion - bad API design on the Git side for not handling the startup case for us.)

_Originally posted by @50Wliu in https://github.com/microsoft/sarif-vscode-extension/pull/522#discussion_r1333610235_

Pybsama:
Closes #525.

## Summary

- initialize the primary workspace repository through the asynchronous `openRepository` API
- await the initialized repository in GitHub analysis and original-document paths
- preserve the existing root-repository selection and submodule bypass behavior
- add offline regression tests for repository selection, initialization, null results, and submodules

## Why

The Git API can discover a repository before its internal model is ready. Returning the object
directly from `git.repositories` leaves startup paths vulnerable to that initialization race.
Waiting for `openRepository` provides the required initialization barrier before repository
methods are used.

## Validation

- `npm test -- --bail false --watch false` (59 passing, 5 pre-existing pending)
- `npm run lint`
- `npm run vscode:prepublish`
- focused repository tests repeated 20 times


Pybsama:
Fresh validation update:

- Upstream `main` is still the original base `bd3eeb13bf155d0d27895d38e0886f427a43fda8`; this PR head is `a26ef62edf0aa699e9c4d3d491c51fc3b627a81f`, ahead 1 / behind 0, with no competing implementation found.
- On Node 18.20.8, the complete suite passed 59/59 with 5 pre-existing pending tests. The complete suite then passed 20 consecutive repetitions.
- ESLint passed, and the production Panel/Context webpack build passed. Generated bundle SHA-256 values: Panel `88d0d014984a8223bcdabf39d0a59ef378ff02dd593e376931eb206e3b91bd61`; Context `7d3eaca6a5753f5a9aaa2d283d003dd20aa25cb86422fc171e44f1686b661894`.
- Node 24.11.1 currently stops before test execution on the unchanged `update.spec.ts` because Mocha cannot import a named export from the old CommonJS Sinon version. The exact upstream base reproduces the same loader failure; it is not introduced by this patch.
- Microsoft CLA is successful. The repository Actions runs named Checks and CodeQL are both `action_required` with zero jobs, so they are waiting for maintainer approval rather than reporting test failures.
- `git diff --check` and the prior independent reviews pass. Ready for review.

